PR #3, opened 2020-01-08
Makes `$runsMigrations` property be always a boolean

nunomaduro:
This pull request makes `$runsMigrations` to be of type `bool` instead of being `bool|null`.


taylorotwell:
Eh, this changes the actual logic of what I was going for.

PR #21, opened 2020-01-10
Fix personal access token expiration

amaelftah:
this **PR** fixes filling **expires_at** column in DB . and also fixes checking of expiration by passing expiration minutes param to **Guard** class


taylorotwell:
That column should not exists. I forgot to remove it.

PR #22, opened 2020-01-10
Add PruneExpiredTokens Schedule command

Te7a-Houdini:
Since Laravel Nova has [PruneStaleAttachments](https://nova.laravel.com/docs/2.0/resources/fields.html#trix-field) . i thought it would be nice idea if we have **PruneExpiredTokens** command

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,6 +181,18 @@ public function boot()
 }
 ```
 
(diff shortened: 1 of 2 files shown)


Te7a-Houdini:
@taylorotwell i will close this since the column is removed

PR #24, opened 2020-01-10
Add AIRLOCK_PREFIX to modify csrf-cookie route prefix

### LINKED ISSUE #17: Unable to change route path
By default, the package loads the route into `/airlock/csrf-cookie`.

Since the application constantly accesses only the `/api` routes, I changed the parameter `baseURL` to `axios.defaults.baseURL = '/api';`

Now when you try to send a GET request to address `/airlock/csrf-cookie`, in fact the request is sent to `/api/airlock/csrf-cookie`.

It would be great to do one of the following options:

1. add the ability to specify the route prefix in file `config/airlock.php`;
2. change the route declaration method similar to `Passport::routes()`.

Te7a-Houdini:
Resolves #17

### DIFF
--- a/config/airlock.php
+++ b/config/airlock.php
@@ -30,4 +30,15 @@ return [
 
     'expiration' => null,
 
+    /*
+    |--------------------------------------------------------------------------
+    | Airlock Prefix
+    |--------------------------------------------------------------------------
+    |
+    | This prefix will be used when hiting initialize CSRF protection route for the application.
+    |
+    */
+
+    'prefix' => env('AIRLOCK_PREFIX'),
+
 ];
--- a/src/AirlockServiceProvider.php
+++ b/src/AirlockServiceProvider.php
@@ -77,10 +77,12 @@ class AirlockServiceProvider extends ServiceProvider
             return;
         }
 
-        Route::get(
-            '/airlock/csrf-cookie',
-            CsrfCookieController::class.'@show'
-        )->middleware('web');
+        Route::prefix(config('airlock.prefix'))->group(function(){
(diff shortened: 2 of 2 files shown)


rennokki:
The config should have a default value?

Te7a-Houdini:
@rennokki i don't think so . cause if no default value presented the route will be without prefix

mbougarne:
Why not to add **prefix** value directly in the config file? ``` 'prefix' => 'api', ``` 
``` php
Route::get(
            '/airlock/csrf-cookie',
            CsrfCookieController::class.'@show'
        )->middleware('web')->prefix(config('airlock.prefix'))
```

Te7a-Houdini:
@mbougarne i just followed [doc example](https://laravel.com/docs/6.x/routing#route-group-prefixes)

meSingh:
@Te7a-Houdini A better implementation will be to have it like this: 

```
/*
|--------------------------------------------------------------------------
| Airlock Prefix
|--------------------------------------------------------------------------
|
| This prefix will be used when hiting initialize CSRF protection route for the application.
|
*/

'route_prefix' => env('AIRLOCK_ROUTE_PREFIX', 'airlock'),
```

```
Route::get(
    config('airlock.prefix') . '/csrf-cookie',
    CsrfCookieController::class.'@show'
)->middleware('web');
```

taylorotwell:
Tweaked this implementation. You can add a prefix configuration option now if you want.

emargareten:
@taylorotwell Is there a reason why there isn't a prefix option in the config file by default?

PR #28, opened 2020-01-11
Improved instruction to publish vendors

TonnyORG:
Minor update to publish airlock's config and migrations.


TonnyORG:
NVM, someone else already fixed this on #27

PR #47, opened 2020-01-17
Add option to select which session driver to use

alexleeelkins:
Currently for serverless applications where the cookie option is not available, subsequent requests using airlock do not work as the session driver is changed to cookie. This allows the user to specify which session driver should be used for subsequent requests.

### DIFF
--- a/config/airlock.php
+++ b/config/airlock.php
@@ -30,15 +30,4 @@ return [
 
     'expiration' => null,
 
-    /*
-     |--------------------------------------------------------------------------
-     | Session Driver
-     |--------------------------------------------------------------------------
-     |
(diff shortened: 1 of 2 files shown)


taylorotwell:
If we were going to do this I'm not sure we would even have a configuration at all? We would just remove where Airlock sets the session entirely and your session would be controlled by config/session.php.

alexleeelkins:
Updated to reflect that. This does put the onus to set the right env variables onto the developer, but removes the current requirement of only using the cookie session driver.

taylorotwell:
We should keep the other two options. I will make the changes.

PR #49, opened 2020-01-19
Modify PersonalAccessToken Model to be polymorphic

### LINKED ISSUE #10: Use polymorphic relationship
Could be useful for issuing tokens for organization accounts.

Te7a-Houdini:
this **PR** resolves #10 . 

by making **PersonalAccessToken** model polymorphic will allow multiple models to be authenticated through airlock tokens .

for example if i have **Pharmacist** & **Doctor** models .  and would authenticate both of them through api  . i would have to [implement this solution](https://github.com/laravel/airlock/issues/10#issuecomment-573019653) which is in my opinion a little bit  tedious .

but by making **PersonalAccessToken** model polymorphic. this will allow implementing the above use case without any extra boilerplate . 

and also customization to user Model through **userModel()** won't be needed any more . cause polymorphic relation will support any other models .

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,7 +175,7 @@ Within your web application's UI, you may wish to list each of the user's tokens
 
 ## Customization
 
-You may customize the "user" model and the personal access token model used by Airlock via the `useUserModel` and `usePersonalAccessTokenModel` methods. Typically, you should call these methods from the `boot` method of your `AppServiceProvider`:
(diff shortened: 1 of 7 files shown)


taylorotwell:
I think I would suggest renaming even the `user` method to something like `tokenable` and then having `tokenable_id` and `tokenable_type` as the morph columns.

Te7a-Houdini:
@taylorotwell Done . and i thnk you are right since the token now belongs to multiple different models .

LearningRaph:
Is there a way to use sanctum without morphing features to save database space and keep a basic user_id column ?

Issue #50, opened 2020-01-20
Airlock::actingAs


leemcd56:
I would really like to use this as an alternative to Passport for an API I am building, but a lot of my tests use `Passport::actingAs` for roles and permissions tests.

How could one go about mocking this functionality for the time being so I could write in tests like this:

```php
   /**
     * Test DELETE /api/properties/{id}.
     *
     * @return void
     */
    public function testDeleteProperty()
    {
        Airlock::actingAs(
            factory(\App\User::class)->create()->attachRole('admin')
        );

        $property = factory(\App\Property::class)->create();
        $response = $this->deleteJson('/api/properties/' . $property->id);

        $response
            ->assertOk()
            ->assertJsonPath('message', 'Deleted property.');

        $this->assertSoftDeleted('properties', [
            'id' => $property->id,
        ]);
    }
```

driesvints:
Feel free to attempt a PR.

Te7a-Houdini:
@leemcd56 i have a quick question . does ``` $this->actingAs($user)``` can solve the case ?

leemcd56:
> @leemcd56 i have a quick question . does ` $this->actingAs($user)` can solve the case ?

@Te7a-Houdini It doesn't create a new personal access token to provide the proper access.

I have opened up PR #51 to add in `::actingAs`. I am having an issue creating a new token because, even though migrations are running, the `personal_access_tokens` table does not appear to exist during the test.

Te7a-Houdini:
Got it thanks for the clarification

driesvints:
PR was merged

PR #51, opened 2020-01-20
Added user mocking through Airlock using actingAs

leemcd56:
This pull request should resolve issue #50 by adding the requested functionality.

## Added

- `Airlock::actingAs` with tests for `auth:api` and abilities

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,13 +192,13 @@ public function boot()
 
 ## Testing
 
-Similar to [Laravel Passport](https://laravel.com/docs/6.x/passport#testing), this method may be used to specify the currently authenticated user.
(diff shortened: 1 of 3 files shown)


taylorotwell:
Both of your tests were broken entirely. Did you run them? I fixed them.

leemcd56:
@taylorotwell I'm sorry about that, I meant to request a review before a merge. The first test worked on my end with no issue, but the second would not. It was saying the `personal_access_tokens` table did not exist, even though the same setup process was used for GuardTest.php. What was I doing wrong?

leemcd56:
It appears to still be failing with the same issue.

16C05C05-A2A0-40D0-9F20-4F96D07630A8

Issue #53, opened 2020-01-21
Unable to exclude URIs from CSRF verification


MatanYadaev:
- Airlock Version: 0.1.0
- Laravel Version: 6.9.0
- PHP Version: 7.4.0
- Database Driver & Version: MySQL 5.7

### Description:
I can't exclude URIs from CSRF verification because of this [LoC](https://github.com/laravel/airlock/blob/master/src/Http/Middleware/EnsureFrontendRequestsAreStateful.php#L30)
Laravel Airlock is using the middleware from the Illuminate package instead of the local one.
This way the `$except` remains always an empty array, and I can't exclude URIs.

### Steps To Reproduce:
1. Add URIs to the `$except` array in `App\Http\Middleware\VerifyCsrfToken.php` file.
2. Try to access this route without a CSRF token.

### Suggestion
Add to the `config/airlock.php` file a configuration variable:
```php
'csrfMiddleware` => App\Http\Middleware\VerifyCsrfToken::class,
```
And make sure to use it in the `EnsureFrontendRequestAreStateful`:
```php
static::fromFrontend($request) ? [
  \Illuminate\Cookie\Middleware\EncryptCookies::class,
  \Illuminate\Cookie\Middleware\AddQueuedCookiesToResponse::class,
  \Illuminate\Session\Middleware\StartSession::class,
  Config::get('airlock.csrfMiddlewareClass', \Illuminate\Foundation\Http\Middleware\VerifyCsrfToken::class),
: []
```

driesvints:
Hmm that indeed seems problematic but I'm not sure if the current solution is the best one? What you can do instead is perhaps bind your implementation to the foundation one through the container:

```
$this->app->bind(
    \Illuminate\Foundation\Http\Middleware\VerifyCsrfToken::class,
    \App\Http\Middleware\VerifyCsrfToken::class
);
```

MatanYadaev:
@driesvints My solution may not be the best one, but binding to the container it's just a workaround in my opinion.
I'm sure more some people are using the `$except` property, so a solution must be documented for them.

lukadriel7:
Hello, I Think the same should apply to the EncryptCookies middleware too.

PR #54, opened 2020-01-21
Add CSRF middleware config

MatanYadaev:
As explained in #53 , using Laravel Airlock is blocking from using a custom `VerifyCsrfToken` middleware.
So I added a `csrfMiddleware` config variable. I'm very bad with the description stuff, can you guys help me with that?

### DIFF
--- a/config/airlock.php
+++ b/config/airlock.php
@@ -30,4 +30,15 @@ return [
 
     'expiration' => null,
 
+    /*
+    |--------------------------------------------------------------------------
+    | CSRF Middleware Class
+    |--------------------------------------------------------------------------
+    |
+    | TBD
+    |
+    */
+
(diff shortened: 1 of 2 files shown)


driesvints:
Maybe it's just best to check if the local class exists and use that one instead of a config option?

```php
$verifyCsrfToken = class_exists(\App\Http\Middleware\VerifyCsrfToken::class)
    ? \App\Http\Middleware\VerifyCsrfToken::class
    : \Illuminate\Foundation\Http\Middleware\VerifyCsrfToken::class;
```

MatanYadaev:
@driesvints I thought about it, but it won't cover cases that this class has a different name.
I guess it's a very edge-case, so maybe it's a better solution, for saving the package as elegant as it is now.

PR #62, opened 2020-01-25
Change README.md

shirazsoft:
Hello,
I think login attempt is much better than manual check.


shahvirag:
Auth::attempt() is not useful for mobile application / third party app authentication, for that genrate airlock auth token.
For SPA you can use Auth::attempt().

shirazsoft:
@shahvirag 
I check laravel docs and i look attemp method do login while we need method to check login certicals.

PR #71, opened 2020-02-12
Add prefix to config

narwy:
The ability to adjust the prefix from the env, most people might want the ability to change it on the fly, default to `airlock`. Similar to `database.php` config for Redis prefix etc, same naming convention. `REDIX_PREFIX` so `AIRLOCK_PREFIX`.


narwy:
This is used in the service provider already: https://github.com/laravel/airlock/blob/af0a5623697e340f7f6587ad3712a19586c82fff/src/AirlockServiceProvider.php#L80

taylorotwell:
You can add it to your configuration file if you want. I don't feel the need to expose it by default as I don't see a big reason to change it.

narwy:
> You can add it to your configuration file if you want. I don't feel the need to expose it by default as I don't see a big reason to change it.

@taylorotwell I would and could see people changing it, even though airlock is a good name. But as a security alternative to now show what kind of auth method you're using, package etc. So in this case you can rename it to `/auth` etc. Gives the option to people out there. If it's listed in the service provider surely it should be in the config by default just like `database` etc? You could say the same about half of the other configs that come with Laravel?

jacobzlogar:
I don't really understand what overhead adding this prefix config would add to the package 🤷‍♂

danabrey:
Having config values that are changeable in config files with their default values is only a positive thing for developer experience. Putting some in there and not others is a source of confusion.

zack6849:
I don't see the issue with adding the documentation for it to the config, why hide away a feature someone might want, there's no additional impact here, just documenting an existing feature in-case someone might want to change it, beats the alternative of someone wanting to change the route prefix and wasting time looking through the code only to find it was configurable from the start and just not documented.

elbojoloco:
Thank you! I've been looking everywhere to find a solution for this problem, never realised it was located and used in the Service Provider. I also can't think of a reason to NOT expose this config option...

narwy:
@taylorotwell Maybe we can get this merged, as there's a demand for it ^

alexleeelkins:
since the prefix configuration is something that is being used, I see no reason to not make the prefix configuration exposed in the configuration file

narwy:
@driesvints Can you check this out please.

imakeinternet:
It is really embarrassing how difficult it has been to get the message across on this SIMPLE request.

imakeinternet:
It is really embarrassing how difficult is has been to get this SIMPLE request baked in.

PR #81, opened 2020-03-05
[1.x] Allow localhost ip access by default

migueldamaso:
Currently the airlock middleware blocks requests from 127.0.0.1 by default only allowing the use localhost itself. This may lead to confusion to certain users especially since if you use artisan serve it recommends the use of 127.0.0.1:8000.

### DIFF
--- a/config/airlock.php
+++ b/config/airlock.php
@@ -13,7 +13,7 @@ return [
     |
     */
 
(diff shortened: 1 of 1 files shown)


narwy:
This should be `0.0.0.0` by default.

PR #84, opened 2020-03-05
Use the API Guard to get the Authenticated User

simonmsara:
Currently I can authenticate successfully by sending a get request to '/airlock/csrf-cookie' the logging in the normal way, but subsequent requests are returning a 401 response with a message of "Unauthenticated". My debugging has led to this change – switching the guard from 'web' to 'api' is authenticated user.

### DIFF
--- a/src/Guard.php
+++ b/src/Guard.php
@@ -42,7 +42,7 @@ class Guard
      */
     public function __invoke(Request $request)
     {
-        if ($user = $this->auth->guard('web')->user()) {
+        if ($user = $this->auth->guard('api')->user()) {
             return $this->supportsTokens($user)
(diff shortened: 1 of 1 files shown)


taylorotwell:
Your app is not configured properly.

phiter:
@taylorotwell do you know which misconfiguration can cause this? I'm trying to get airlock working for two days unsuccessfully. There is no stack trace, so I have no idea where I could be wrong.

simonmsara:
@phiter I think it's the default guard in the auth config, it has to be web.
```config('auth.defaults.guard')```
Hard coding it to 'api' won't help either, maybe we should pass the 'web' guard as the default but still use what the configured guard is.

PR #86, opened 2020-03-06
[1.x] Using the Default Guard Defined in the Auth Config

simonmsara:
Currently the default guard can be changed but we hard code the web guard in this case, changing it to a dynamic definition of the name still allows us to change the config. If this poses some issues we need to update the documentation and state that the default must be set to 'web'.


taylorotwell:
Probably needs its own separate Airlock configuration item.

PR #92, opened 2020-03-10
Add support for all subdomains

adrum:
There is currently not a way to accept all subdomains when setting `AIRLOCK_STATEFUL_DOMAINS `. We are building an application where cookies on the root domain should work on all subdomains. This PR changes the logic to allow for that scenario.

I considered making it a config option to allow subdomains, but I could not think of any negative consequences by making this the default behavior.

### DIFF
--- a/src/Http/Middleware/EnsureFrontendRequestsAreStateful.php
+++ b/src/Http/Middleware/EnsureFrontendRequestsAreStateful.php
@@ -54,11 +54,11 @@ class EnsureFrontendRequestsAreStateful
      */
     public static function fromFrontend($request)
     {
-        $referer = Str::replaceFirst('https://', '', $request->headers->get('referer'));
+        $url = \parse_url($request->headers->get('referer'));
 
-        $referer = Str::replaceFirst('http://', '', $referer);
+        $referer = $url['host'];
 
-        return Str::startsWith(
+        return Str::endsWith(
             $referer,
             config('airlock.stateful', [])
         );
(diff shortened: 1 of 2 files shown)


taylorotwell:
Worried about this change from startsWith to endsWith leading to subtle breaking changes. Personally I would maybe try changing things to parse the host like you have now but using `Str::is`?

adrum:
@taylorotwell I may need some help coming up with an appropriate pattern, since my first thought would be to do something like this:

```
return Str::is(
             "*$referer",
             config('airlock.stateful', [])
         );
```

which, is pretty much the same thing as endsWith.

Also, another reason I chose this direction is because, in theory, someone could have their `airlock.stateful` set to something like `laravel.com` and then a third party could write an application using a domain such as `laravel.com.thirdparty.com`, which airlock would allow as a first party domain by using the `startsWith` approach as it exists today. Using endsWith closes that hole.

adrum:
Another approach might be to map over the values in `airlock.stateful` and do an exact match if the value doesn't start with a `.`. If it does, then we could technically use the `endsWith` approach, which would probably behave as people expect.

Example:

`laravel.com` does not match subdomains.
`.laravel.com` does match subdomains.

taylorotwell:
@adrum the whole first party request thing isn't really security sensitive. It doesn't matter if a third party tries to trick it.

taylorotwell:
Fixed and tagged 1.0.1... you can now use * as a wildcard.

emargareten:
Since it's using `Str::is` and checking the referrer (including the path) you would need to specify `*laravel.com*`, not just `*laravel.com`.
@taylorotwell is there a reason why you chose `referrer` header instead of `origin`?

emargareten:
I think [this commit](https://github.com/laravel/airlock/commit/70450c3583d3a0cfa3d61c4ffda88820d4f93f24) was fine, just had to add an asterisk after `$referrer`.

PR #96, opened 2020-03-12
Wildcard Stateful Config

liamkeily:
I have a case where whitelisting specific domains isn't practical. For example services like [Zeit](https://zeit.co/) deploy development branches with unique `xxxxxxxx.company.now.sh` domain names. It would be nice if there was a way to define a wildcard.


adrum:
This was just added two hours ago: https://github.com/laravel/airlock/releases/tag/v1.0.1

PR #97, opened 2020-03-12
SameSite Cookie Changes in February 2020

AndrewSavetchuk:
We can't set the value to `lax` here, because of the new changes to the way cookies are handled in Chrome. 

If the `same_site` is set to 'lax' requests don't work in Chrome 80+, it says `401 Unauthorized`. It should be set to `none` in `config/session.php`.

Please, see links below: 
https://blog.chromium.org/2020/02/samesite-cookie-changes-in-february.html
https://github.com/laravel/airlock/issues/11#issuecomment-596438725
https://github.com/laravel/airlock/issues/11#issuecomment-598266966

### DIFF
--- a/src/Http/Middleware/EnsureFrontendRequestsAreStateful.php
+++ b/src/Http/Middleware/EnsureFrontendRequestsAreStateful.php
@@ -42,7 +42,6 @@ class EnsureFrontendRequestsAreStateful
     {
         config([
             'session.http_only' => true,
-            'session.same_site' => 'lax',
         ]);
     }
 


taylorotwell:
Nope, it should be set to `lax` for security. Your unauthorized problem is something else.

taylorotwell:
Also, do not use `localhost` when testing Airlock. Use real subdomains with something like Valet.

taylorotwell:
I think many people are running into issues because they setup two applications both on localhost but on different ports... those are not considered "same site" requests and cookies will not send even with same site `lax`... so people are setting to `none` thinking they solved their problem.

IMO it's important to use real subdomains, not `localhost`.

genesiscz:
@taylorotwell so turns out Chrome wants me to use SameSite=none 

image

Testing frontend on localhost.com:2021
sanctum.php having localhost.com:2021 in stateful
API is on whole different domain (which is probably the problem, but there's no way around that now)

How do I fix that?

aliakbarazizi:
> Nope, it should be set to `lax` for security. Your unauthorized problem is something else.

the problem is when you set config to `lax` you overwrite default behavior, I set config to `None` and ok sanctum can use `lax` but why you change my config in framework? it took me 6 hours to find and I was surprised that Sanctum actually modifying my config.

PR #110, opened 2020-03-22
[2.x] Make the guard configurable

bambamboole:
I am building a Laravel package which uses sanctum. It was really a missing peace in the eco system. Thanks for this!

However, since I have an additional guard in my package, I had problems to make the described stateful requests.

There was also no way to swap out the implementation via the container because it is instantiated with `new` instead of using the container.

This little change would give me the possibility to go on with this package. I think there is also a discussion about this topic in #20  

There are no breaking changes.

### DIFF
--- a/config/sanctum.php
+++ b/config/sanctum.php
@@ -17,19 +17,6 @@ return [
 
     /*
     |--------------------------------------------------------------------------
-    | Guard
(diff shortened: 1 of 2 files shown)


iSWORD:
This is very helpful and needs to be added to the config file.

PR #118, opened 2020-03-28
Add auth guard to sanctum configs

mostafakram:
Added guard to sanctum configs to allow developers to define custom auth guards 

It will not break existing features as the `Guard` class already get the user by `$this->auth->guard(config('sanctum.guard', 'web'))->user()`

I have copied how nova config file define the guard and replaced 'NOVA' with 'SANCTUM' so that it uses the same standards as laravel packages

### DIFF
--- a/config/sanctum.php
+++ b/config/sanctum.php
@@ -42,8 +42,7 @@ return [
     'middleware' => [
         'verify_csrf_token' => App\Http\Middleware\VerifyCsrfToken::class,
     ],
-
-
+    
     /*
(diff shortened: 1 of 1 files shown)


driesvints:
This was removed from the original pr. Feel free to add this in your own app.

mostafakram:
How should developers know that this is the way to define another guard if its not in congis and not in documentation ?

msonowal:
Hi @driesvints  why this has been removed this is a good config and required where people used different guards and need to use that with sanctum

Issue #119, opened 2020-04-02
Adding guard variable to sanctum.php


ibadin:
I suggest to place 'guard' variable in sanctum.php

clugg:
See #118

PR #128, opened 2020-04-15
Add config option that is already added to the code

gauravmak:


clugg:
See #118 and #119

gauravmak:
Okay, I tried to search and found that this was removed from the [original PR](https://github.com/laravel/sanctum/pull/110/commits/42fb21056b27268b1f60b1e5797ea0bbcd59abb8) but do not understand the reasoning behind this.

Agree that maintainers cannot answer each and everyone but having it written somewhere will save contributors' as well as maintainers' time in the future.

PR #131, opened 2020-04-18
Add app url domain to stateful

MammutAlex:
Problem: when user create SPA in one project (local + app), sanctum auth cannot work on app because domain APP_URL is not defined in stateful. So good idea is to add SANCTUM_STATEFUL_DOMAINS into .env.

In that pull I added APP_URL domain into stateful to solve that.

### DIFF
--- a/config/sanctum.php
+++ b/config/sanctum.php
@@ -13,7 +13,7 @@ return [
     |
     */
 
-    'stateful' => explode(',', env('SANCTUM_STATEFUL_DOMAINS', 'localhost,127.0.0.1')),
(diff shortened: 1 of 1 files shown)


clugg:
This is the point of environment variables - if you need to add more domains (i.e. your APP_URL), you can add them in your `.env` file. What is stopping you from doing that?

MammutAlex:
@clugg Nothing, but I think better to add APP_URL by default to don’t do any additional actions for typical projects and avoid problems when everything works well on local server but after deploy on app server stops

PR #132, opened 2020-04-19
[2.x] Use origin instead of referer

sausin:
Good case is made for use of origin instead of referer [here](https://www.sjoerdlangkemper.nl/2019/02/27/prevent-csrf-with-the-origin-http-request-header/). As mentioned, some AV software strip out the referer in requests, which might lead to API requests failing.

The origin header doesn't contain the full path of the referring domain. If, however, there is a need to use the referer for convenience it can be specified as an option to the user in the sanctum config. I can add commits to make this possible if that's considered a better way forward.


driesvints:
This seems like quite a big breaking change, no?

PR #138, opened 2020-05-07
[2.x] Fix wildcard subdomain issue

tecbay:
At this moment the `EnsureFrontendRequestsAreStateful@fromFrontend` method is basically checking the `referer` header which is somehow works for none wildcard domain , But it isn't works in wildcard subdomain #6 . Because the `referer` header contain the full path url . I changed it to `origin` header check . Which is work perfectly for wildcard subdomain and also for the reguler domain.
Screenshot 2020-05-07 at 11 33 05 PM


driesvints:
This has been been tried before. See https://github.com/laravel/sanctum/pull/132

PR #145, opened 2020-05-21
[2.x] Optionally disable routes.

ellisio:
We are using Sanctum for API tokens and Native apps only. Because of this, we don't need to expose the csrf-cookie route. My OCD is not happy with a route being listed that is never being used. This adds the ability to disable the route.


taylorotwell:
You can now add `'routes' => false` to your Sanctum configuration file.

ellisio:
Thanks @taylorotwell! 🍻

PR #149, opened 2020-06-03
[2.x] Added Multiple Provider Support

RahulDey12:
So far Laravel Sanctum has not supported any multiple providers based auth guard. So I have added a function called `hasValidProvider` which accepts the `tokenable` model and then it matches if the `tokenable_type` model is an `instanceof` the provider type. if no provider defined it always returns `true`.

You need to add guards like this:
```
'guards' => [
        'web' => [
            'driver' => 'session',
            'provider' => 'users',
        ],

        'api' => [
            'driver' => 'sanctum',
            'provider' => 'users',
        ],

        'admin' => [
            'driver' => 'session',
            'provider' => 'admins',
        ],

        'admin-api' => [
            'driver' => 'sanctum',
            'provider' => 'admins',
        ],
    ],
```

> And you can also use the guard `sanctum` as before it will work fine.


src/Guard.php
```diff
@@ -25,13 +32,15 @@ class Guard
      * Create a new guard instance.
      *
      * @param  \Illuminate\Contracts\Auth\Factory  $auth
+     * @param string $provider
      * @param  int  $expiration
      * @return void
      */
-    public function __construct(AuthFactory $auth, $expiration = null)
+    public function __construct(AuthFactory $auth, $provider, $expiration = null)
```
driesvints, line 39: We can't do this breaking change on a stable release.

RahulDey12: Tonight i'll fix that

RahulDey12: Done

KoenLav:
This PR seems to break our usage of "auth:sanctum" as a Guard for GraphQL calls (using Lighthouse).

Tried to pinpoint what's causing it, but haven't been able to find out.

We've downgraded to 2.3.3 and all is fine now.

RahulDey12:
@KoenLav How ?

itkarkkainen:
Just noticed the same problem as KoenLav. Confirmed by downgrading from 2.4.0 to 2.3.3. For unknown reason axios calls fail when auth:sanctum route guard is added.

`Route::middleware('auth:sanctum')->group(function() { ... }`

ivand88:
This is exactly the thing I was looking for. I'd just wanted to add that in config/sanctum.php you should set:

`'guard' => null`

Otherwise in your session based SPA you could get response code 200 for the resources where you expect it to be 401.

reza305z:
@ivand88 points to a real problem, but the solution he provides is wrong.
That is because if one sets the guard value to null in sanctum config file, then sanctum driver will not have any session-based guards to get the authenticated user.

reza305z:
I wonder if laravel developers have any plan to add multi-auth to sanctum for both spa and api token authentication?

ivand88:
> @ivand88 Or instead of using web guard, you can define a custom guard with session driver.

It's not that simple. Please look at the source for the sanctum, and the whole discussion within this PR. If you add a custom guard in the sanctum config, you will still get a false 200 response code for some resources, when you don't actually expect it.
If needed, I can elaborate more.

reza305z:
> > @ivand88 Or instead of using web guard, you can define a custom guard with session driver.
> 
> It's not that simple. Please look at the source for the sanctum, and the whole discussion within this PR (Especially issue which mentions this PR #232). If you add a custom guard in the sanctum config, you will still get a false 200 response code for some resources, when you don't actually expect it. If needed, I can elaborate more.

I'd be glad if you explain more.

dietergeerts:
> So far Laravel Sanctum has not supported any multiple providers based auth guard. So I have added a function called `hasValidProvider` which accepts the `tokenable` model and then it matches if the `tokenable_type` model is an `instanceof` the provider type. if no provider defined it always returns `true`.
> 
> You need to add guards like this:
> 
> ```
> 'guards' => [
>         'web' => [
>             'driver' => 'session',
>             'provider' => 'users',
>         ],
> 
>         'api' => [
>             'driver' => 'sanctum',
>             'provider' => 'users',
>         ],
> 
>         'admin' => [
>             'driver' => 'session',
>             'provider' => 'admins',
>         ],
> 
>         'admin-api' => [
>             'driver' => 'sanctum',
>             'provider' => 'admins',
>         ],
>     ],
> ```
> 
> > And you can also use the guard `sanctum` as before it will work fine.
> 
> > Note: If you want to use multi auth in SPA that case you should use session guards like `auth:admin,admin-api`

Based on this, I was expecting to use something like

```
   Route::group(['middleware' => ['auth:admin-api']], function () {
    });
```

But that doesn't seem to work, because when I want to login with `Auth::guard('admin-api`)->login`, I get errors that `login` doesn't exist? Is there any more complete example of how this is supposed to work?

RahulDey12:
`login` method does not exist in Sanctum Guard if you want to login any user please use the `admin` guard on SPA and refer this for token creation https://laravel.com/docs/10.x/sanctum#issuing-api-tokens

RahulDey12:
@dietergeerts It's actually possible with laravel you just need to call `Auth::guard('admin')->login()` & `Auth::guard('web')->login()`

For the routing it is
```php
Route::group(['middleware' => ['auth:admin-api', 'auth:admin']], function () {
    });
```

Issue #144, opened 2020-05-21
Dynamic guard change for spa authorization


sergey-yabloncev:
If the project uses multiple guards, for example web and admin, you can't use spa authorization for multiple guards. 

This feature is provided for authorization via token.

https://github.com/laravel/sanctum/blob/2.x/src/Guard.php#L45
https://github.com/laravel/sanctum/blob/2.x/src/SanctumServiceProvider.php#L95

I would like to be able to set a guard in middleware, to use a similar design

```php
Route::middleware(['auth:sanctum'])->get('/user', function () {
    return auth('web')->user();
});

// and in one project

Route::middleware('auth:sanctum')->get('/admin', function () {
    return auth('admin')->user();
});
```

driesvints:
If you want to implement something like this just send in a pr. Thanks.

lnpbk:
@sergey-yabloncev did you manage to implement this, or begin working on anything?

sergey-yabloncev:
> @sergey-yabloncev did you manage to implement this, or begin working on anything?

@lnpbk unfortunately, no. I refused to use the package as a SPA Authentication, I used it as the Tokens API. Additionally, by adding custom middleware checking the instanceof model of the authorized user

for example

```php

<?php

namespace App\Http\Middleware;

use App\Models\User\Admin;
use Closure;

class VerifyAdmin
{
    /**
     * Handle an incoming request.
     *
     * @param \Illuminate\Http\Request $request
     *
     * @return mixed
     */
    public function handle($request, Closure $next)
    {
        if (!(auth()->user() instanceof Admin)) {
            abort(403);
        }

        return $next($request);
    }
}
```
App\Http\Kernel.php

```php
<?php

namespace App\Http;
//...
use App\Http\Middleware\VerifyAdmin;
use App\Http\Middleware\VerifyUser;

class Kernel extends HttpKernel
{
//...
    protected $routeMiddleware = [
        //...
        'admin' => VerifyAdmin::class,
        'user' => VerifyUser::class,
    ];
}
```
use in route or controller

```php
Route::middleware(['auth:sanctum', 'admin'])
```

IlCallo:
If you accept to avoid config caching, you can use a middleware to change it at runtime.

```php
// app/Http/Middleware/UpdateSanctumConfigForCentralApp.php

<?php

namespace App\Http\Middleware;

use Closure;

class UpdateSanctumConfigForCentralApp
{
    /**
     * Handle an incoming request.
     *
[...]

PR #154, opened 2020-06-06
[2.x] Sanctum Guard now falls back to default driver

lnpbk:
I'm running into an issue where I need to use Sanctum for multiple session-based auth guards. This simple change means that I can use `auth()->setDefaultDriver('something-else')` before the `auth:sanctum` middleware is run, and have my guard recognised.

This issue was raised [here](https://github.com/laravel/sanctum/issues/144)

### DIFF
--- a/src/Guard.php
+++ b/src/Guard.php
@@ -51,7 +51,7 @@ class Guard
      */
     public function __invoke(Request $request)
     {
-        if ($user = $this->auth->guard(config('sanctum.guard', $this->auth->getDefaultDriver()))->user()) {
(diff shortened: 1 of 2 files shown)


lnpbk:
It seems I got a little ahead of myself, and didn't realise that the `Guard` expects an instance of `AuthFactory`. I've tweaked the code, and added a test

taylorotwell:
No plans to change this atm.

lnpbk:
@taylorotwell @driesvints Are there plans to support multiples guards in the future? It’d be great to get some more clarity on the rejection.

driesvints:
@lnpbk we have no plans to add support for that at this time.

PR #168, opened 2020-07-08
Added guard to config file

markvesterskov:
After having spent a solid two hours trying to figure out why i could not get Sanctum to authenticate after a successful login, it dawned on me that it was using the wrong guard for session validation. 

Custom guards are already implemented, but not mentioned in the documentation nor config file, so I've added it to the config file for people to find.


taylorotwell:
Doesn't fit rest of configuration formatting. I will add personally.

PR #174, opened 2020-08-09
Remove port from referer to check for hostname

KoenLav:
Our application broke when upgrading to 2.4.x due to the (accidentally?) introduced change of regarding the port to be a part of the domain.

It's debatable whether port should be included, so it's definitely also an option to not merge this, but then maybe we should add some explanation somewhere (or at least make mentioning of this in the changelog).


driesvints:
Heya, it was actually always expected that the port was included. The previous behavior wasn't actually meant to work. The docs have been updated since:

Screenshot 2020-08-09 at 17 18 18

PR #179, opened 2020-08-19
get default guard name from config

masihjay-z:
ّIn some applications may not have web guard and it caused to face to "Auth guard [web] is not defined." error.
It is better to try to get default guard from config ("auth.defaults.guard"), if "auth.defaults.guard" didn't present then hard code it to "web".


msonowal (approved):

taylorotwell:
This is already customizable... Use `sanctum.guard`.

PR #183, opened 2020-08-24
PersonalAccessToken::findToken can also return `null`

canvural:
PersonalAccessToken::findToken can also return `null`. This PR adds it to the return doc block.


driesvints:
Thanks

PR #194, opened 2020-09-14
[3.x] Enhancement/registers routes

KaniRobinson:
### Hey guys,

> Just an opener to say, I absolutely love what you guys have done with Laravel 8! Sanctum, Fortify, JetStream, Livewire, and Inertia are amazing and compliment the development process in Laravel so much!

#### Whats is the reasoning behind my PR?

I wanted to disable the default routes that come with JetStream, Fortify, and Sanctum so my thought process was, let's have a look to see if there is any code in the RouteServiceProvider to enable/disable the routes, similar to the old `Auth::routes()`, nope...

I then began to dig deeper and look into the Service Provider of each vendor package and found the winning code to do this: 
`Fortify::$registersRoutes = false;` and `Jetstream::$registersRoutes = false;`. This left me with just Sanctum, which to my knowledge was actually set via the `sanctum.php` config file.

So my thought behind this PR was to just keep the flow of the codebases as similar as possible and include, in the same format as Fortify and JetStream just so people can have 1 standard to follow instead of config for one package and a variable for another :) 

Please excuse me if this is a waffle of a PR and I'd really like to hear whether you'd prefer to stick down this route of the `registersRoutes` variable or move all packages over to use the `config` 😁

### DIFF
--- a/src/Sanctum.php
+++ b/src/Sanctum.php
@@ -21,6 +21,13 @@ class Sanctum
     public static $runsMigrations = true;
 
     /**
+     * Indicates if Sanctum routes will be registered.
+     *
+     * @var bool
+     */
+    public static $registersRoutes = true;
+
+    /**
      * Set the current user for the application with the given abilities.
      *
      * @param  \Illuminate\Contracts\Auth\Authenticatable|\Laravel\Sanctum\HasApiTokens  $user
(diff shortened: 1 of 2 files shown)


KaniRobinson:
@driesvints All moved to master branch 👍

taylorotwell:
I would just this as-is. We can add a "routes" config entry for the other two.

KaniRobinson:
@taylorotwell Either way it'll be a breaking change, how comes you'd prefer config over variables? Caching?

Would you like me to push a pull request for JetStream & Fortify?

PR #199, opened 2020-09-25
[2.x] Allow table name customization

tontonsb:


driesvints:
You can already use `Sanctum::usePersonalAccessTokenModel($model)` to personalize the access token and overwrite the `getTable` method so I don't think we need another config option.

BuyOwnEx:
If I use Sanctum::usePersonalAccessTokenModel($model), I get the following error in NewAccessToken.php:31
TypeError
Argument 1 passed to Laravel\Sanctum\NewAccessToken::__construct() must be an instance of Laravel\Sanctum\PersonalAccessToken, instance of App\PersonalAccessToken given

My code in Controller the following:

public function token($client_id)
    {
        Sanctum::usePersonalAccessTokenModel('App\PersonalAccessToken');
        $user = Client::find($client_id);
        $token = $user->createToken('token-name');
        return $token->plainTextToken;
    }

BuyOwnEx:
I solved this problem via creating own class NewAccessToken

PR #204, opened 2020-10-09
[2.x] Adds origin header fallback

### LINKED ISSUE #203: Checking for Origin instead of (or as well as) Referer
I've had a long-standing issue with Sanctum that I've finally solved today. Every time I used it I kept getting authorization errors despite following everything in the Docs and reading countless Stack Overflow posts.

Most solutions always revolve around `SANCTUM_STATEFUL_DOMAINS` not being configured correctly. I knew mine were ok.

I've noticed in the past that using a different browser from my default (Firefox) actually works. I've put this down to a weird Firefox bug that would get fixed in the next update. But it never did.

I finally traced the issue to `EnsureFrontendRequestsAreStateful::fromFrontend()` which checks the referer against the stateful domains. It turns out Firefox was not sending the `referer` header. Sure enough in the settings the `network.http.sendRefererHeader` option was set to `0` (off).

Turns out that a VPN Firefox extension (Private Internet Access) decided to turn this off for me as a security measure. It also does the same in the Chrome version.

So it turns out that anyone using these extensions (or any that does the same thing) will not be able to use a website that utilizes Sanctum.

I have noticed the `Origin` header has always been present in the request. Is it possible to also check this header as a backup in case `Referer` does not exist? Or are there issues with the `Origin` header that I'm not aware of?

dellow:
Adds an `origin` fallback if the `referer` has been stripped. Closes #203.

### DIFF
--- a/src/Http/Middleware/EnsureFrontendRequestsAreStateful.php
+++ b/src/Http/Middleware/EnsureFrontendRequestsAreStateful.php
@@ -55,7 +55,7 @@ class EnsureFrontendRequestsAreStateful
      */
     public static function fromFrontend($request)
     {
-        $domain = ($request->headers->get('referer')) ? $request->headers->get('referer') : $request->headers->get('origin');
+        $domain = $request->headers->get('referer') ?: $request->headers->get('origin');
         $domain = Str::replaceFirst('https://', '', $domain);
         $domain = Str::replaceFirst('http://', '', $domain);
         $domain = Str::endsWith($domain, '/') ? $domain : "{$domain}/";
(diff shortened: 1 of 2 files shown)


src/Http/Middleware/EnsureFrontendRequestsAreStateful.php
```diff
@@ -55,14 +55,15 @@ protected function configureSecureCookieSessions()
      */
     public static function fromFrontend($request)
     {
-        $referer = Str::replaceFirst('https://', '', $request->headers->get('referer'));
-        $referer = Str::replaceFirst('http://', '', $referer);
-        $referer = Str::endsWith($referer, '/') ? $referer : "{$referer}/";
+        $domain = ($request->headers->get('referer')) ? $request->headers->get('referer') : $request->headers->get('origin');
```
driesvints, line 58:
```suggestion
        $domain = $request->headers->get('referer') ? $request->headers->get('referer') : $request->headers->get('origin');
```

driesvints, line 58:
```suggestion
        $domain = $request->headers->get('referer') ?: $request->headers->get('origin');
```

PR #216, opened 2020-11-05
[2.x] Return correct CSRF Token

Komar12124:


driesvints:
Please resubmit with a thorough explanation why this is needed.

PR #217, opened 2020-11-05
[2.x] Return correct CSRF Token 

Komar12124:
First we need add cookies to response then start the session and only after that encrypt cookies.
Otherwise, while requesting for new cookies we will get the error: CSRF token mismatch

### DIFF
--- a/src/Http/Middleware/EnsureFrontendRequestsAreStateful.php
+++ b/src/Http/Middleware/EnsureFrontendRequestsAreStateful.php
@@ -20,14 +20,14 @@ class EnsureFrontendRequestsAreStateful
         $this->configureSecureCookieSessions();
 
         return (new Pipeline(app()))->send($request)->through(static::fromFrontend($request) ? [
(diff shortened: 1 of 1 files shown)


taylorotwell:
This is incorrect.

Komar12124:
> This is incorrect.

Sorry, my bad. Not attentive :)

Komar12124:
@taylorotwell @driesvints Any way, i dont really understand.
```
    /**
     * Determine if the session and input CSRF tokens match.
     *
     * @param  \Illuminate\Http\Request  $request
     * @return bool
     */
    protected function tokensMatch($request)
    {
        $token = $this->getTokenFromRequest($request);

        Log::debug([$request->session()->token(), $token]);

        return is_string($request->session()->token()) &&
               is_string($token) &&
               hash_equals($request->session()->token(), $token);
    }
```

Why in my sequence of pipes tokens matching fine, but in default they are different?

PR #225, opened 2020-11-22
[2.x] Fix user provider in `sanctum` guard

sebdesign:
This PR provides the specified user provider to the `sanctum` guard.

Without that, the `Illuminate\Auth\RequestGuard` is constructed without a user provider by default, which is inconsistent with the `Laravel\Sanctum\Guard` callable which accepts the specified user provider.

I've added tests to prove that the `sanctum` guard is not properly configured:

The `Laravel\Sanctum\Guard` callable accepts the specified provider in the constructor from the configuration, and validates if the provider model matches the authenticated model. This part is working correctly.

In the service provider, the actual guard `Illuminate\Auth\RequestGuard` is constructed with the default application provider which is `null`, instead of the specified provider in the configuration.

So, calling `Auth::guard('sanctum')->user()` or `Auth::guard('sanctum')->validate()` will succeed because it's using the user provider configured in the service provider.

But `Auth::guard('sanctum')->getProvider()` will return `null`. With this fix it will return the correct user provider.

### DIFF
--- a/src/SanctumServiceProvider.php
+++ b/src/SanctumServiceProvider.php
@@ -114,7 +114,7 @@ class SanctumServiceProvider extends ServiceProvider
         return new RequestGuard(
             new Guard($auth, config('sanctum.expiration'), $config['provider']),
             $this->app['request'],
-            $auth->createUserProvider()
+            $auth->createUserProvider($config['provider'])
         );
     }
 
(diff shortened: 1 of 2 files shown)


sebdesign:
@taylorotwell thanks for merging this. IMHO the multiple provider feature (#149) is a bit flaky. Right now the `RequestGuard` accepts and instance of the user provider but the `Guard` accepts the name of the provider. But the `Guard` callback is always invoked with the user provider instance as a second parameter, so there is no need to inject it in the constructor.

Also only the `EloquentUserProvider` has as `model` key in the config. The `DatabaseUserProvider` (or any custom provider) wouldn't have this key.

I think in this should be refactored. Would you accept a PR in 2.x or should I wait for 3.x?

taylorotwell:
I dunno. I don't really put a lot of thought into using things outside of Eloquent 🤷 

If something needs to be refactored feel free to submit a PR.

PR #227, opened 2020-11-23
Change bigIncrements('id') to id() in migration

Xinecraft:
Changed id column in  `2019_12_14_000001_create_personal_access_tokens_table.php`

I am not sure why in all Jetstream scaffolding this is the only place which is not changed to id(), I am not sure maybe someone can explain if it got a reason, 
if not then I think it's good to be consistent. 😅


Xinecraft:
Sorry forgot about backward compatibility I am a noob 😅

PR #236, opened 2020-12-23
Added Sanctum Default Guard

thecaliskan:
Here you may specify which authentication guard Sanctum will use whileauthenticating users. This value should correspond with one of your guards that is already present in your "auth" configuration file.

### DIFF
--- a/config/sanctum.php
+++ b/config/sanctum.php
@@ -20,6 +20,19 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Sanctum Guard
+    |--------------------------------------------------------------------------
+    |
+    | Here you may specify which authentication guard Sanctum will use while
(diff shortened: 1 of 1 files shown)


taylorotwell:
It should pretty much just always use the `web` guard. I don't really want to confuse people even more than they already are with this auth stuff. If you need this option you can add it to your own configuration file.

thecaliskan:
Hi;

I use this system with Nova. Two guards are required in this scenario. This is not confusion.  If there is a possibility of define, it should be. No time should be wasted to find.

thecaliskan:
Needs to be standard in config files on Laravel Official Projects.

thecaliskan:
@driesvints what do you think about this subject ?

PR #240, opened 2021-01-05
add sanctum:cleanup command

masihjay-z:


driesvints:
Thanks. There's no breaking change here so you can send this to 2.x if you like.

PR #243, opened 2021-01-12
Add APP_URL to default sateful domain list

grzegorz-rozycki:
Currently when setting up sanctum on a new project the project domain has to be added to the stateful list.

A nice-to-have feature would be to also add the domain based on the APP_URL environment variable.


driesvints:
This seems to break the test suite.

PR #245, opened 2021-01-21
[2.x] Add support multiple guard for SPA auth

stevenbdf:
This pull request adds support to multiple guards for SPA Authentication.

By default it will take the `sanctum.guard` value, if not it will get all the guards set in `auth.guards` that the drivers are session.

Iterates all the valid `guards` and returns the first item that matches.

### DIFF
--- a/src/Guard.php
+++ b/src/Guard.php
@@ -51,10 +51,18 @@ class Guard
      */
     public function __invoke(Request $request)
     {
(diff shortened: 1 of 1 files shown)


taylorotwell:
Spinning through all configured guards sounds like a bad idea. Better than sanctum.guard just allows an array.

stevenbdf:
Can I proceed to implement this multi guard support making `sanctum.guard` an array ?

PR #247, opened 2021-01-24
[2.x] Return json response when the request expects a json

nmfzone:
There is known issue with Laravel Sanctum in Postman (https://github.com/postmanlabs/postman-app-support/issues/2418#issuecomment-645771459).

If `sanctum/csrf-cookie` support returning JsonResponse, somehow it fixed those issue. So, when a Postman user sending request to the `sanctum/csrf-cookie`, they need to pass `Accept` header as `application/json` to make it working.

### DIFF
--- a/src/Http/Controllers/CsrfCookieController.php
+++ b/src/Http/Controllers/CsrfCookieController.php
@@ -2,6 +2,8 @@
 
 namespace Laravel\Sanctum\Http\Controllers;
 
+use Illuminate\Http\JsonResponse;
+use Illuminate\Http\Request;
 use Illuminate\Http\Response;
 
 class CsrfCookieController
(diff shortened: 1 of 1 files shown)


cortier:
Not working. Tested with Postman Version 11.19.0.
```

Content-Type: application/json
Accept: application/json
Cache-Control: no-cache
Postman-Token: 97a1b492-7c62-4b4d-8038-f488a753e620
```

Still getting:
`Error: Parse Error: Expected HTTP/`

cortierdev:
If the problem persists, check with curl that no extra bytes are being sent from Laravel:

```
Excess found: excess = 4 url = / (zero-length body)
```

No bytes should be included on a 204 response.

PR #252, opened 2021-02-23
[3.x] Add expires_at functionality to tokens.

Zoefschildpad:
Adds an expires_at timestamp to tokens. If the expires_at has passed, the token can no longer be used. 

This may allow users to set different expiration times for different tokens. So you can create a token used for a one-time import that expires in automatically in a week, and a token that generates daily reports that doesn't expire.

This also lets developers enforce different expiration times for different tokens based on the type of user, or the privileges of the token.

### DIFF
--- a/database/migrations/2019_12_14_000001_create_personal_access_tokens_table.php
+++ b/database/migrations/2019_12_14_000001_create_personal_access_tokens_table.php
@@ -20,6 +20,7 @@ class CreatePersonalAccessTokensTable extends Migration
             $table->string('token', 64)->unique();
             $table->text('abilities')->nullable();
             $table->timestamp('last_used_at')->nullable();
+            $table->timestamp('expires_at')->nullable();
             $table->timestamps();
         });
     }
--- a/src/Guard.php
+++ b/src/Guard.php
@@ -65,6 +65,8 @@ class Guard
             if (! $accessToken ||
                 ($this->expiration &&
                  $accessToken->created_at->lte(now()->subMinutes($this->expiration))) ||
+                ($accessToken->expires_at &&
+                    $accessToken->expires_at->isPast()) ||
                 ! $this->hasValidProvider($accessToken->tokenable)) {
                 return;
             }
--- a/src/HasApiTokens.php
+++ b/src/HasApiTokens.php
@@ -2,6 +2,7 @@
 
 namespace Laravel\Sanctum;
 
+use Illuminate\Support\Carbon;
 use Illuminate\Support\Str;
 
 trait HasApiTokens
@@ -39,14 +40,16 @@ trait HasApiTokens
      *
      * @param  string  $name
      * @param  array  $abilities
+     * @param  Carbon|null $expires_at
      * @return \Laravel\Sanctum\NewAccessToken
      */
-    public function createToken(string $name, array $abilities = ['*'])
+    public function createToken(string $name, array $abilities = ['*'], Carbon $expires_at = null)
     {
         $token = $this->tokens()->create([
             'name' => $name,
             'token' => hash('sha256', $plainTextToken = Str::random(40)),
             'abilities' => $abilities,
+            'expires_at' => $expires_at,
         ]);
 
         return new NewAccessToken($token, $token->id.'|'.$plainTextToken);
--- a/src/PersonalAccessToken.php
+++ b/src/PersonalAccessToken.php
@@ -27,7 +27,7 @@ class PersonalAccessToken extends Model implements HasAbilities
         'name',
         'token',
         'abilities',
-        'expires_at'
+        'expires_at',
     ];
 
     /**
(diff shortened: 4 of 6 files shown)


ankurk91:
It has been 1 year and this feature has not been released.

Any plans for v3?

driesvints:
Yeah I have it on my list to look into. Maybe in Q2.

lindyhopchris:
Just noting that the `Carbon $expires_at = null` argument on the `createToken` method should be type-hinted as `CarbonInterface` instead, for apps that have set `CarbonImmutable` as their default date class.

driesvints:
@lindyhopchris it was changed to DateTimeInterface in a later commit.

lindyhopchris:
@driesvints ah great! sorry, just looked at the PR diff.

ankurk91:
This date should be considered in Model prune command?

qschmick:
Just highlight that this updates an existing migration so applications with this already installed will not add the new field. This should have been a new migration to add the field.

ankurk91:
There must be an upgrade guide in repo

tcampbPPU:
Do I need to manually add the migration to that table myself if I am currently using version: `2.15`?

sfneal:
Here's the migration I created to add an 'expires_at' column when upgrading from v2 to v3

```php
<?php

use Illuminate\Database\Migrations\Migration;
use Illuminate\Database\Schema\Blueprint;
use Illuminate\Support\Facades\Schema;

return new class extends Migration
{
    /**
     * Run the migrations.
     *
     * @return void
     */
    public function up()
    {
        Schema::table('personal_access_tokens', function (Blueprint $table) {
            $table->timestamp('expires_at')->after('last_used_at')->nullable();
        });
    }

    /**
     * Reverse the migrations.
     *
     * @return void
     */
    public function down()
    {
        Schema::table('personal_access_tokens', function (Blueprint $table) {
            $table->dropColumn('expires_at');
        });
    }
};
```

qschmick:
Good one @sfneal just FYI for all that see this, you can't add this migration to a v3 version as that column will already exists and the migration will fail. If you make the following change it could be added without an issue:

```php
<?php

use Illuminate\Database\Migrations\Migration;
use Illuminate\Database\Schema\Blueprint;
use Illuminate\Support\Facades\Schema;

return new class extends Migration
{
    /**
     * Run the migrations.
     *
     * @return void
     */
    public function up()
    {
        if (! Schema::hasColumn('personal_access_tokens', 'expires_at') {
            Schema::table('personal_access_tokens', function (Blueprint $table) {
                $table->timestamp('expires_at')->after('last_used_at')->nullable();
            });
        }
    }

    /**
     * Reverse the migrations.
     *
     * @return void
     */
    public function down()
    {
        Schema::table('personal_access_tokens', function (Blueprint $table) {
            $table->dropColumn('expires_at');
        });
    }
};
```

Apfelwurm:
@qschmick youre migration is nice, but it's currently missing a closing bracket in the if clause.

fixed one:

```
<?php

use Illuminate\Database\Migrations\Migration;
use Illuminate\Database\Schema\Blueprint;
use Illuminate\Support\Facades\Schema;

return new class extends Migration
{
    /**
     * Run the migrations.
     *
     * @return void
     */
    public function up()
    {
        if (! Schema::hasColumn('personal_access_tokens', 'expires_at')) {
            Schema::table('personal_access_tokens', function (Blueprint $table) {
                $table->timestamp('expires_at')->after('last_used_at')->nullable();
            });
        }
    }

    /**
     * Reverse the migrations.
     *
     * @return void
     */
    public function down()
    {
        Schema::table('personal_access_tokens', function (Blueprint $table) {
            $table->dropColumn('expires_at');
        });
    }
};
```

qschmick:
Thanks @Apfelwurm great catch

dan-the-dev:
> Here's the migration I created to add an 'expires_at' column when upgrading from v2 to v3
> 
> ```
> <?php
> 
> use Illuminate\Database\Migrations\Migration;
> use Illuminate\Database\Schema\Blueprint;
> use Illuminate\Support\Facades\Schema;
> 
> return new class extends Migration
> {
>     /**
>      * Run the migrations.
>      *
>      * @return void
>      */
>     public function up()
>     {
>         Schema::table('personal_access_tokens', function (Blueprint $table) {
>             $table->timestamp('expires_at')->after('last_used_at')->nullable();
>         });
>     }
> 
>     /**
>      * Reverse the migrations.
>      *
>      * @return void
>      */
>     public function down()
>     {
>         Schema::table('personal_access_tokens', function (Blueprint $table) {
>             $table->dropColumn('expires_at');
>         });
>     }
> };
> ```

This should be more exposed and suggested as migration step from 2 to 3 in official places.

qschmick:
Does appear to be in the upgrade.md

https://github.com/laravel/sanctum/blob/3.x/UPGRADE.md

PR #264, opened 2021-03-26
[2.x] Environment APP_URL added into the default sanctum.stateful configuration

polakjan:
## What?

Added the application's URL from `.env`'s `APP_URL` to the default value of `sanctum.stateful` configuration.

## Why?

In my experience (teaching new Laravel developers) not having the application's URL among default stateful domains leads to a lot of headache when deploying to a production environment. The URL changes from e.g. `localhost:3000` to `www.test.com` and suddenly a new configuration value must be added into `.env` which was not needed until now.

While everyone expects they should change the `APP_URL` when moving to a new environment, few also remember to add `SANCTUM_STATEFUL_DOMAINS`

This is also hard to notice because the application works, only the requests which should be authenticated are not.

## How?

Added the host part of the `APP_URL` environment value to `/config/sanctum.php` 

## Testing

I needed to modify the `phpunit.xml.dist` file to provide a sample env `APP_URL` configuration that could be loaded with the `env()` function.

Tests test the presence of the host of `env('APP_URL')` in `config('sanctum.stateful'`) and that a request coming from the environment `APP_URL` configuration passes the `Laravel\Sanctum\Http\Middleware\EnsureFrontendRequestsAreStateful::fromFrontend()` check.

## Duplicite PR?

This may seem like a duplicite of #243 but does the same thing with much simpler code and the test suite passes.

### DIFF
--- a/config/sanctum.php
+++ b/config/sanctum.php
@@ -15,7 +15,7 @@ return [
 
     'stateful' => explode(',', env(
         'SANCTUM_STATEFUL_DOMAINS',
-        'localhost,localhost:3000,127.0.0.1,127.0.0.1:8000,::1,' . parse_url(env('APP_URL'), PHP_URL_HOST)
+        'localhost,localhost:3000,127.0.0.1,127.0.0.1:8000,::1,'.parse_url(env('APP_URL'), PHP_URL_HOST)
     )),
 
     /*
--- a/config/sanctum.php
+++ b/config/sanctum.php
@@ -15,7 +15,7 @@ return [
 
     'stateful' => explode(',', env(
         'SANCTUM_STATEFUL_DOMAINS',
-        'localhost,localhost:3000,127.0.0.1,127.0.0.1:8000,::1'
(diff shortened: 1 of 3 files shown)


driesvints:
This shouldn't be needed I believe. Why can't you just set `SANCTUM_STATEFUL_DOMAINS`? Why is the change to the phpunit config file needed?

polakjan:
You can just set `SANCTUM_STATEFUL_DOMAINS`. This is more a convenience thing. The default stateful configuration already contains "traditional", common values like `localhost:3000` while the most common value - the `APP_URL` set up in `.env` is not there.

When you forget to set the `SANCTUM_STATEFUL_DOMAINS` configuration which you don't need for local development, you don't get any warning that something is wrong, only SPA requests fail to be authenticated. Hours of head scratching. This simple adjustment could help avoid that.

The change to phpunit config file was needed to provide a sample a sample env APP_URL configuration that could be loaded with the env() function to test if it appears in sanctum's `stateful` configuration. I used information from `orchestral/testbench` Is there any other way to write that test?

PR #270, opened 2021-04-16
[2.x] Add HasApiTokens contract to complement trait

wimski:
In a recent project using Sanctum I wanted to check if the `User`/`Model`/`Authenticatable`/whatever uses the `HasApiTokens` trait. Currently this can only be achieved by one of the following options:

```php
// A. check trait usage
if (! in_array(HasApiTokens::class, class_uses($user)) {
    throw new Exception();
}

// B. check method existence
if (! method_exists($user, 'currentAccessToken')) {
    throw new Exception();
}
```

* Option A will work, but static analysis will not understand that `$user` has a specific set of methods.
* Option B does support static analysis, but is messy, because it doesn't really check for the trait (or specific contract).

Therefore I've added a contract to complement the trait.

```php
use Laravel\Sanctum\Contracts\HasApiTokens as HasApiTokensContract;

if (! $user instanceof HasApiTokensContract::class) {
    throw new Exception();
}
```

This contract is not required for general use of Sanctum and is not a breaking change.

### DIFF
--- a/src/Contracts/HasApiTokens.php
+++ b/src/Contracts/HasApiTokens.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 namespace Laravel\Sanctum\Contracts;
 
 interface HasApiTokens
@@ -41,7 +39,7 @@ interface HasApiTokens
      * Set the current access token for the user.
      *
      * @param  \Laravel\Sanctum\Contracts\HasAbilities  $accessToken
-     * @return HasApiTokens
+     * @return \Laravel\Sanctum\Contracts\HasApiTokens
      */
     public function withAccessToken($accessToken);
 }
--- a/src/Contracts/HasApiTokens.php
+++ b/src/Contracts/HasApiTokens.php
@@ -1,0 +1,47 @@
+<?php
+
+declare(strict_types=1);
+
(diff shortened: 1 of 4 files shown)


src/Contracts/HasApiTokens.php
```diff
@@ -0,0 +1,47 @@
...
+     *
+     * @return \Laravel\Sanctum\Contracts\HasAbilities
+     */
+    public function currentAccessToken();
+
+    /**
+     * Set the current access token for the user.
+     *
+     * @param  \Laravel\Sanctum\Contracts\HasAbilities  $accessToken
+     * @return HasApiTokens
```
driesvints, line 44: This can't be a trait. Trait isn't a valid type to return in DocBlocks as they're not objects or interfaces.

driesvints: Sorry, realized this is an interface of course. Needs to be an FQN:

```suggestion
     * @return \Laravel\Sanctum\Contracts\HasApiTokens
```

wimski: It doesn't return a trait. That is the contract (itself) with the same name.

wimski:
Thanks for merging!

When can I expect the next release containing this merge?

PR #279, opened 2021-05-19
Only parse `APP_URL` for default stateful domains when it's set

dshoreman:
#### Problem
If `APP_URL` is *not* defined in the .env, the config breaks on PHP 8.1 with a null string deprecation. Since it has to load in the service provider, artisan and composer are broken too (not sure about new installs though—I only have 8.1 via GH actions).

For older versions, it's less of an issue - it merely adds `''` to the `sanctum.stateful` array. Looks scary from a distance, but there's an `array_filter()` in the middleware so it's harmless... I think.

**Introduced in v2.9.3** via #264.

#### Notes
A simple `env('APP_URL', '')` would suffice for the 8.1 issue, but it still leaves the extra empty entry in `sanctum.stateful`. To remove any chance of worry/risk/ambiguity, this PR makes the whole thing conditional on `APP_URL` being defined.

### DIFF
--- a/config/sanctum.php
+++ b/config/sanctum.php
@@ -13,10 +13,11 @@ return [
     |
     */
 
-    'stateful' => explode(',', env(
-        'SANCTUM_STATEFUL_DOMAINS',
-        'localhost,localhost:3000,127.0.0.1,127.0.0.1:8000,::1,'.parse_url(env('APP_URL'), PHP_URL_HOST)
-    )),
+    'stateful' => explode(',', env('SANCTUM_STATEFUL_DOMAINS', sprintf(
+        '%s%s',
+        'localhost,localhost:3000,127.0.0.1,127.0.0.1:8000,::1',
+        env('APP_URL') ? ','.parse_url(env('APP_URL'), PHP_URL_HOST) : ''
+    ))),
 
     /*
     |--------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


config/sanctum.php
```diff
@@ -13,10 +13,11 @@
...
     */
 
-    'stateful' => explode(',', env(
-        'SANCTUM_STATEFUL_DOMAINS',
-        'localhost,localhost:3000,127.0.0.1,127.0.0.1:8000,::1,'.parse_url(env('APP_URL'), PHP_URL_HOST)
-    )),
+    'stateful' => explode(',', env('SANCTUM_STATEFUL_DOMAINS', sprintf(
+        '%s%s',
+        'localhost,localhost:3000,127.0.0.1,127.0.0.1:8000,::1',
+        env('APP_URL') ? ','.parse_url(env('APP_URL'), PHP_URL_HOST) : ''
```
dshoreman, line 19: Initially tried a simple concat here, but the nesting (IMO) makes it much harder to see what's happening at a glance:
```php
'stateful' => explode(',', env(
    'SANCTUM_STATEFUL_DOMAINS',
    'localhost,localhost:3000,127.0.0.1,127.0.0.1:8000,::1'.(
        env('APP_URL') ? ','.parse_url(env('APP_URL'), PHP_URL_HOST) : ''
    )
));
```

I realise not everybody's a fan of `sprintf('%s%s', ...)` though so I can change this up if needed

PR #283, opened 2021-06-08
Ignore updating `last_used_at` for deciding the DB connection host

alexjose:
**Problem**
Sanctum updates the access token's `last_used_at` property every time an API is authenticated using sanctum guard. This db query happens in the initial phase of the request cycle. This causes all the subsequent db queries to go to writer host even in the case of read operations.

**Solution**
Saving the updated `last_used_at` operation is moved to the application terminating event.

### DIFF
--- a/src/Guard.php
+++ b/src/Guard.php
@@ -70,7 +70,7 @@ class Guard
             }
 
             return $this->supportsTokens($accessToken->tokenable) ? $accessToken->tokenable->withAccessToken(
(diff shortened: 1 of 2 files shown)


taylorotwell:
Hmm - I think you have a good point but we need to solve it in a different way. We can't use the `app` function here. I will look into it.

taylorotwell:
I have updated this PR - does it look OK to you?

alexjose:
Yes. This approach looks much better.

PR #286, opened 2021-06-13
updated app()['request'] with request()

RahulDey12:
Using the request helper instead of `app()['request']`


driesvints:
Why?

RahulDey12:
> Why?

It was looking weird.

PR #295, opened 2021-07-13
Add configuration options to the configuration file

countless-integers:
I've noticed some configuration look-ups in project classes that were not in the published configuration file. I'd have saved me some time if they were there. Not sure if that's on purpose, but seeing how all the other options are there even if defaults are set in look-ups, I thought it makes sense to have them all in there.

TBH I might not have enough flair to write those nice PHP docs, so I'd love some help with those.


driesvints:
Please see previous attempts.

PR #307, opened 2021-10-08
fix: replace hardcoded "web" guard by config('sanctum.guard')

StephaneBour:
Replace hardcoded "web" guard by config('sanctum.guard') (same as here : https://github.com/laravel/sanctum/blob/2.x/src/Guard.php#L55 )


driesvints:
We're reverting this: https://github.com/laravel/sanctum/pull/309

PR #320, opened 2021-11-12
Access token Id inconsistent in default migration

darthsoup:
The token ID does not match the ID in the laravel/laravel project.

It should not be a breaking change. 

See: https://github.com/laravel/laravel/blob/8.x/database/migrations/2019_12_14_000001_create_personal_access_tokens_table.php


darthsoup:
Yikes, I forgot that Sanctum still supports Laravel 6, which doesn't have `$table->id()`.

PR #321, opened 2021-11-15
[2.x] Added Personal access token format checking

### LINKED ISSUE #318: An incorrectly formed bearer token causes SQL error
- Sanctum Version: 2.12.1
- Laravel Version: 8.70.2
- PHP Version: 8.0.9
- Database Driver & Version: PostgreSQL 13.2

### Description:
An incorrectly formed token (a token with a quote or double quote on the left) causes SQL error on the server. We discovered it as a result of a mistake made by our frontend engineer.

The line causing the error:
https://github.com/laravel/sanctum/blob/2.x/src/PersonalAccessToken.php#L64

>Invalid text representation: 7 ERROR:  invalid input syntax for type bigint: ""138" (SQL: select * from "personal_access_tokens" where "personal_access_tokens"."id" = "138 limit 1) {"exception":"[object] (Illuminate\\Database\\QueryException(code: 22P02): SQLSTATE[22P02]: Invalid text representation: 7 ERROR:  invalid input syntax for type bigint: \"\"138\" (SQL: select * from \"personal_access_tokens\" where \"personal_access_tokens\".\"id\" = \"138 limit 1) at /var/www/project/releases/20211110162512/vendor/laravel/framework/src/Illuminate/Database/Connection.php:703)

An example of an incorrectly formed token: (Note the double quotes on the left.)
`"138|E5bndnZUVCQUVkYWphbUV2SmlGcHg1UWFZeG0yRGd`

**I don't know if this is a bug or something that I should be dealing with.** We fixed the part that caused the token to be corrupted on our part, but I wanted to let you know because the 500 error gives me the feeling that something is wrong.

### Steps To Reproduce:

```
curl --request POST \
  --url https://project.test/api/broadcasting/auth \
  --header 'Accept: application/json' \
  --header 'Authorization: Bearer "138|E5bndnZUVCQUVkYWphbUV2SmlGcHg1UWFZeG0yRGd' \
  --header 'Content-Type: application/json'
```

RahulDey12:
Added extra checking for Personal Access Token format.

Note: This issue was not coming to the MySQL & I don't have any PgSQL set up so I will request @mgsmus please check if the issue still coming.

closes #318

### DIFF
--- a/src/PersonalAccessToken.php
+++ b/src/PersonalAccessToken.php
@@ -61,9 +61,11 @@ class PersonalAccessToken extends Model implements HasAbilities
 
         [$id, $token] = explode('|', $token, 2);
 
-        if ($instance = static::find($id)) {
+        if (is_numeric($id) && $instance = static::find($id)) {
             return hash_equals($instance->token, hash('sha256', $token)) ? $instance : null;
         }
+
+        return null;
     }
 
     /**
--- a/tests/GuardTest.php
+++ b/tests/GuardTest.php
@@ -270,6 +270,29 @@ class GuardTest extends TestCase
         $this->assertNull($user);
     }
 
+    public function test_is_token_properly_formatted()
+    {
+        $this->loadLaravelMigrations(['--database' => 'testbench']);
+        $this->artisan('migrate', ['--database' => 'testbench'])->run();
+
+        $user = User::forceCreate([
+            'name' => 'Taylor Otwell',
+            'email' => 'taylor@laravel.com',
+            'password' => '$2y$10$92IXUNpkjO0rOQ5byMi.Ye4oKoEa3Ro9llC/.og/at2.uheWG/igi',
+            'remember_token' => Str::random(10),
+        ]);
+
+        $token = PersonalAccessToken::forceCreate([
+            'tokenable_id' => $user->id,
+            'tokenable_type' => get_class($user),
+            'name' => 'Test',
+            'token' => hash('sha256', 'test'),
+        ]);
+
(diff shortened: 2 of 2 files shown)


taylorotwell:
What if the user is using UUID primary keys?

src/PersonalAccessToken.php
```diff
@@ -61,9 +61,11 @@ public static function findToken($token)
 
         [$id, $token] = explode('|', $token, 2);
 
-        if ($instance = static::find($id)) {
+        if (is_numeric($id) && $instance = static::find($id)) {
```
mgsmus, line 64: Since the primary key can also be a string, I think it will be better if we do it like this:

```
[$id, $token] = explode('|', $token, 2);

if(static::getKeyType() === 'int' && !is_int($id)) {
    return null;
}

if ($instance = static::find($id)) {
    return hash_equals($instance->token, hash('sha256', $token)) ? $instance : null;
}
```
@driesvints what do you think?

taylorotwell:
This whole error is not even really something that can typically happen in user land to be honest. It was a typo on a developer's part, so I'm not sure we need to try to put all sorts of checks around it.

src/PersonalAccessToken.php
RahulDey12: You cannot call `getKeyType` statically

RahulDey12:
> This whole error is not even really something that can typically happen in user land to be honest. It was a typo on a developer's part, so I'm not sure we need to try to put all sorts of checks around it.

I agree but, it would be better if we just wrap a try-catch around the `if` statement and show unauthenticated on any kind of exception

Issue #325, opened 2021-12-08
token retrieval as dependency injection


sonirico:
It would be great if the retrieval of the token payload could be configurable as a closure receiving the request defaulting to the bearer aproach.

driesvints:
Think it's best that you propose something through a PR so we have a better idea what you mean. Thanks

PR #326, opened 2021-12-08
feat: add custom fetch token option #325

sonirico:
The rationale behind this commit is to offer a configurable way to obtain the sanctum token from the request so as to broaden the functionality of this package to other scenarios. In my case, this package would fit very well for webhooks authentication and authorization. However, webhooks as you know can't be expected to have any kind of headers.

@driesvints This should be a draft of what I mean. In essence it makes configurable how the token is retrieved from the request. Tests need to be added and your opinion would be much appreciated as I am not PHP/Laravel expert.

About the PR, perhaps should be interesting to make somehow the retrieval of the token configurable on per route basis?

### DIFF
--- a/config/sanctum.php
+++ b/config/sanctum.php
@@ -62,4 +62,25 @@ return [
         'encrypt_cookies' => App\Http\Middleware\EncryptCookies::class,
     ],
 
+
+    /*
+    |--------------------------------------------------------------------------
+    | Custom fetch token
+    |--------------------------------------------------------------------------
+    |
+    | Leaving this as `null` will default to fetching the token from the
+    | `Authorization` header as bearer tokens. If your token should be in other
+    | parts of the request, you may change this to a function
+    | receiving the request:
+    |
+    |  ```php
+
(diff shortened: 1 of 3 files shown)


driesvints:
You can't place a function call in a config file. It's also best to explain in your PR what this PR does so Taylor understands why this is needed.

sonirico:
> You can't place a function call in a config file. It's also best to explain in your PR what this PR does so Taylor understands why this is needed.

The function call *is not* is the config file, only its declaration so as to be used later in the package. I have added a little more context in the description of this PR.

driesvints:
@sonirico your example in the description block above is showing the usage which literally shows the function in the config file.

taylorotwell:
If there another way you want to retrieve the token just PR it to the guard.

sonirico:
> If there another way you want to retrieve the token just PR it to the guard.

hi @taylorotwell , can you provide with more guidance about how to proceed? As I said I am new to the community. Which guard do you mean?

Issue #304, opened 2021-09-28
Sanctum does not send authentication lifecycle events


alexking:
- Sanctum Version: 2.11.1
- Laravel Version: 8.45.1
- PHP Version: 7.4.21
- Database Driver & Version: mysqlnd 7.4.21

### Description:
Sanctum does not send [authentication lifecycle events](https://laravel.com/docs/8.x/authentication#events). Not all of the events listed in that documentation make sense for non-stateful authentication (there's no real login or logout for example), but whenever the guard is invoked and a valid token is found, a `Illuminate\Auth\Events\Authenticated` event should be dispatched for consistency. Would be happy to make a fix for this. 

### Steps To Reproduce:

To reproduce be sure you're using tokens - all events will send fine when using the SessionGuard. When you authenticate with a token, no events will be sent.

driesvints:
This is more of a feature request. You're always free to attempt a PR to see if Taylor would accept it 👍

PR #327, opened 2021-12-14
[2.x] Add an event on successful token validation

aalyusuf:
This PR solve cases like #304 and #315 without breaking changes


taylorotwell:
Renamed to `TokenAuthenticated`.

PR #333, opened 2022-01-27
[2.x] Configurable default token abilities

billriess:
This PR adds the ability to configure the default abilities new tokens receive. This is achieved with a new `sanctum.default_abilities` config key. If the config is not present it will fall back to the default behavior of `['*']` so that this is not a breaking change.

If the config is already published, adding this key to the existing `sanctum.php` will allow you to configure the default abilities new tokens receive.

### DIFF
--- a/src/HasApiTokens.php
+++ b/src/HasApiTokens.php
@@ -41,12 +41,12 @@ trait HasApiTokens
      * @param  array  $abilities
      * @return \Laravel\Sanctum\NewAccessToken
      */
-    public function createToken(string $name, array $abilities = ['*'])
+    public function createToken(string $name, array $abilities = null)
     {
         $token = $this->tokens()->create([
             'name' => $name,
             'token' => hash('sha256', $plainTextToken = Str::random(40)),
-            'abilities' => $abilities,
+            'abilities' => $abilities ?? config('sanctum.default_abilities', ['*']),
         ]);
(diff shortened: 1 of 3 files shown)


src/HasApiTokens.php
```diff
@@ -41,12 +41,12 @@ public function tokenCan(string $ability)
      * @param  array  $abilities
      * @return \Laravel\Sanctum\NewAccessToken
      */
-    public function createToken(string $name, array $abilities = ['*'])
+    public function createToken(string $name, array $abilities = null)
```
driesvints, line 44: This is unfortunately a breaking change for anyone overriding this method.

driesvints:
I believe it's best to target master (3.x) for these changes.

taylorotwell:

billriess:
Should I target `master` with this PR instead? While this change does make minor changes to the interface and trait, it allows the package to be more flexible in a fairly straightforward way. In many cases, it's dangerous to assume a user wants to grant all abilities to a new token via `tokenCreate()` when no abilities are specified. This PR simply allows you to configure this behavior.

driesvints:
@billriess I'm sorry but it seems Taylor doesn't wants to implement this. I suggest to build a predefined list of default abilities in your app and pass that with the tokenCreate method.

PR #339, opened 2022-02-12
Switch to anonymous migrations

willemo:
I noticed Sanctum wasn't using anonymous migrations, while other parts of the ecosystem are. This PR changes the migration to be anonymous.

I've asked on the Discord server if 2.x is the correct branch for this PR, but I haven't had any reply yet. If this is not the correct branch, I can make a new PR for the correct one.


taylorotwell:
Seems to be breaking our test build. 🤔

PR #340, opened 2022-02-12
Automatically prepend the port onto the APP_URL when using the fallback for valid stateful domains

mikemellor11:
Not including the local port is a [common issue](https://stackoverflow.com/questions/61421547/getting-401-unauthorized-for-laravel-sanctum) when setting up Sanctum auth for new users.

There's already config code to fallback to the APP_URL if the SANCTUM_STATEFUL_DOMAINS env var isn't present - this just sures up that fallback code to also include the port if its preset and ignores it if not.


taylorotwell:
Cleaned this up a bit: https://github.com/laravel/sanctum/commit/57023176c5a77d5108cee2fcadceabda00af814b

Thanks

PR #343, opened 2022-02-16
[2.x] Revert extracting url buildup

driesvints:
Reverting this because this won't work with config caching. See https://github.com/laravel/sanctum/commit/57023176c5a77d5108cee2fcadceabda00af814b#commitcomment-66786793


taylorotwell:
No need to revert - doesn't break config caching - as person noted it would only not work if you called it outside config files - which I do not intend for users to do. Regardless, I will change it to use `config('app.url')`

PR #344, opened 2022-02-22
Use `$this->id()` instead of `$table->bigIncrements('id')` in migration stub.

dyanakiev:
By default, in the [laravel/laravel](https://github.com/laravel/laravel/blob/9.x/database/migrations/2019_12_14_000001_create_personal_access_tokens_table.php#L17) package it's using `$this->id()`, but if i vendor:publish the sanctum package (or install jetstream) the line is overwritten with `$table->bigIncrements('id')` I believe it's better to use `$this->id()` here too.


dyanakiev:
I did not expect tests to be failing on older PHP versions `BadMethodCallException: Method Illuminate\Database\Schema\Blueprint::id does not exist.`

driesvints:
Older Laravel versions don't have this method.

PR #348, opened 2022-03-23
[2.x] Add sanctum:prune-expired command for removing expired tokens.

yuraplohov:
This command has long been needed by developers.
A similar command is present in Telescope and Passport packages.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,6 +16,7 @@
     "require": {
         "php": "^7.2|^8.0",
         "ext-json": "*",
+        "illuminate/console": "^6.9|^7.0|^8.0|^9.0",
         "illuminate/contracts": "^6.9|^7.0|^8.0|^9.0",
         "illuminate/database": "^6.9|^7.0|^8.0|^9.0",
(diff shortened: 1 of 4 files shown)


ankurk91:
The command does not return any thing.
The exit codes must be added

driesvints:
Thanks @ankurk91. I sent in a PR for that: https://github.com/laravel/sanctum/pull/351

EdgarSedov:
@yuraplohov @ankurk91 @driesvints 
Hey guys, thanks for awesome command, it's good that we don't need to write it ourselves.

I have one question - is there any special reason why we use `created_at` instead of `last_used_at`?

yuraplohov:
Thank you, @EdgarSedov 

I used `created_at` in this logic, because expiration of token in Sanctum based on `created_at` exactly. Maybe, expiration by `last_used_at` is needed in this package, but that's another question. I saw PR [#349](https://github.com/laravel/sanctum/pull/349) with this idea recently.

Issue #361, opened 2022-04-28
how to revoke all tokens except the current token


HAstaShakiba:
in sanctum, we can revoke all or the current token.
Is there a way to revoke all tokens except the current one?

driesvints:
Hi there,

Thanks for reporting but it looks like this is a question which can be asked on a support channel. Please only use this issue tracker for reporting bugs with the library itself. If you have a question on how to use functionality provided by this repo you can try one of the following channels:

- [Laracasts Forums](https://laracasts.com/discuss)
- [Laravel.io Forums](https://laravel.io/forum)
- [StackOverflow](https://stackoverflow.com/questions/tagged/laravel)
- [Discord](https://discordapp.com/invite/KxwQuKb)
- [Larachat](https://larachat.co)
- [IRC](https://web.libera.chat/?nick=laravelnewbie&channels=#laravel)

However, this issue will not be locked and everyone is still free to discuss solutions to your problem!

Thanks.

HAstaShakiba:
I looked for the answer but did not find it.
A simple way:

`        $currentAccessToken = $request->user()->currentAccessToken();
`

`         auth()->user()->tokens()->whereNot('id', $currentAccessToken->id)->delete();
`

But can this not be an internal function?

`$user->otherAccessToken()->delete();`

driesvints:
@HAstaShakiba you can always attempt a PR if you want 👍

PR #362, opened 2022-04-28
revoke all tokens except the current access token

HAstaShakiba:
#361


driesvints:
Seems like something went wrong with branching here.

PR #380, opened 2022-07-22
Adding ability to add csrf cookie route name

bencarter78:
This PR gives developers the opportunity to add a route name for the csrf-cookie route. When working with an SPA in tandem with something like https://github.com/tighten/ziggy, you can pass route names to a route helper function. This will keep it consistent with being able to name other routes in a standard Laravel application.

This adds a new item in the sanctum config which is used in the route name, if it is null it is ignored.

### DIFF
--- a/config/sanctum.php
+++ b/config/sanctum.php
@@ -64,5 +64,5 @@ return [
         'encrypt_cookies' => App\Http\Middleware\EncryptCookies::class,
     ],
 
-    'sanctum.cookie_route_name' => env('SANCTUM_COOKIE_ROUTE_NAME')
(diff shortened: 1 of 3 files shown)


taylorotwell:
Why not just assign it a name instead of making a configuration option?

bencarter78:
@taylorotwell Do you mean there's already a way to add a name to a defined route or that this PR should just scrap the configuration option?

taylorotwell:
I mean there shouldn't be a configuration option. Just assign it a name. I don't know why it would need to be configurable.

taylorotwell:
Added this route name on 3.x.

PR #385, opened 2022-07-28
Prune expires_at tokens

### LINKED ISSUE #384: Prune command doesn't take expires_at into account
- Sanctum Version: 3.0.0
- Laravel Version: #.#.#
- PHP Version: #.#.#
- Database Driver & Version:

### Description:
As mentioned here https://github.com/laravel/sanctum/pull/252#issuecomment-1198125142 the `sanctum:prune-expired` does not prune models based on their expires_at column, but only on the global `sanctum.expiration` config setting. This is counterintuitive and incorrect. The prune command should be updated for v3.

iruoy:
Closes #384

This still requires the `sanctum.expiration` to be set.

### DIFF
--- a/src/Console/Commands/PruneExpired.php
+++ b/src/Console/Commands/PruneExpired.php
@@ -39,7 +39,7 @@ class PruneExpired extends Command
 
         if ($expiration = config('sanctum.expiration')) {
             $this->components->task(
-                'Pruning tokens with expired expiration value',
+                'Pruning tokens with expired expiration value based on configuration file',
                 fn () => $model::where('created_at', '<', now()->subMinutes($expiration + ($hours * 60)))->delete()
             );
         } else {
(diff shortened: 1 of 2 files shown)


driesvints:
Actually I was thinking that we might be able to do this without requiring `sanctum.expiration` to be set? Maybe we should always allow pruning by `expired_at` and if `sanctum.expiration is set additionally prune by `created_at`?

iruoy:
That is how I envisioned it before looking at the code. However I didn't want to make the change before getting some feedback. I'll look into it later today.

iruoy:
I've pushed some changes. Tokens with `expired_at` will now always be pruned and I've used the new artisan `task()` component.

I've kept the `warn()` component for now, but I think it should be removed.

driesvints:
@iruoy you're sure that tokens without an expired_at timestamp won't be deleted, right?

PR #386, opened 2022-07-29
Update migration's primary identifier change

suyar:
This is the migration file in laravel：
`2019_12_14_000001_create_personal_access_tokens_table.php`
```
....
Schema::create('personal_access_tokens', function (Blueprint $table) {
    $table->id();
...
```

This is the migration file in sanctum：
`2019_12_14_000001_create_personal_access_tokens_table.php`
```
...
Schema::create('personal_access_tokens', function (Blueprint $table) {
    $table->bigIncrements('id');
...
```

I think the id field should be kept consistent.


driesvints:
I think we couldn't do this before with Sanctum supporting older versions but seems to be okay now.

driesvints (approved):

ankurk91:
Can we also switch to using anonymous class in migration?

PR #390, opened 2022-08-13
Add feature able to custom token length in config file

afiqiqmal:
Currently the default token length which is 40 and unable to customize without extend the method. So, this pull request intended to make the sanctum token length customizable in `HasApiTokens` trait via sanctum config file.

I added token length value in sanctum.php config file and also added test to check the token length


.gitignore
```diff
@@ -2,3 +2,4 @@
 composer.lock
 /phpunit.xml
 .phpunit.result.cache
+.idea
```
driesvints, line 5: This belongs in a global gitignore file. 
```suggestion
```

taylorotwell:

PR #395, opened 2022-08-30
add migration file for adding expires_at field

mohamedgaber-intake40:
add migration file for adding expires_at field
when upgrading to v3 the record for 2019_12_14_000001_create_personal_access_tokens_table already exists in the migrations table in the DB so it will not add the new column and ignore the modifications for the original migration file


driesvints:
This isn't needed. You need to create a new migration yourself: https://github.com/laravel/sanctum/blob/3.x/UPGRADE.md#new-expires_at-column

mohamedgaber-intake40:
may I know why the package doesn't need a new file and the user needs to create the file by himself?
I mean why the package doesn't provide the file for the user and it will work when upgrading or installing the package in a new project?

PR #399, opened 2022-10-01
feature: added config option to disable handling of bearer tokens

dciprian-petrisor:
## Description
This pull request adds the ability to disable the default behaviour of Sanctum to interpret bearer tokens in the request header if none of the provided guards were able to authenticate the request.

## Details
Currently, if you ignore Sanctum's database migrations because you want to rely on, for example, a custom Guard which authenticates with a third party auth service instead, and this Guard fails to authenticate the user (e.g. expired/invalid tokens), then
Sanctum will also attempt to validate the Bearer token by itself, which results in a table not found database error.

In order to prevent such issues, we can prevent Sanctum from handling bearer tokens via a config option.

**This change is fully backwards compatible & keeps the current behaviour, unless specifically set to false in the config.**

### DIFF
--- a/config/sanctum.php
+++ b/config/sanctum.php
@@ -37,6 +37,19 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Bearer Token Handling
+    |--------------------------------------------------------------------------
+    |
+    | This value controls whether Sanctum should use the bearer token that's
+    | present on an incoming request for authentication if none of the Guards
+    | were able to authenticate the it.
+    |
+    */
+
+    'handle_bearer_tokens' => true,
+
+    /*
+    |--------------------------------------------------------------------------
     | Expiration Minutes
     |--------------------------------------------------------------------------
     |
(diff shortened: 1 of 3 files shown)


driesvints:
What's the use of Sanctum if you use a third party auth service?

dciprian-petrisor:
> What's the use of Sanctum if you use a third party auth service?

I use Sanctum for its SPA authentication features only, using session cookies.

dciprian-petrisor:
I will offer a bit more context:

The app where I ran into this scenario uses Keycloak to authenticate users.

Sanctum is used for session cookie, with a custom laravel user provider which reads the user out of Keycloak.

For enabling API access for users, keycloak issues JWTs, which I verify using a custom Sanctum Guard which I registered in 'sanctum.config'.

If this guard does not return a user, then sanctum also attempts to read the token and authenticate it, which I do not want.

Its just occurring to me now that I may be misusing the sanctum.guard config somehow, please let me know if that is the case.

Thanks!

taylorotwell:

PR #402, opened 2022-10-14
add missing expiresAt parameter to contract

simbig:
This pull request adds the missing expiresAt parameter to the contract.


taylorotwell:
Not sure we can change this on a patch release? @driesvints

PR #403, opened 2022-10-19
refactor: early return was used to keep readability

saeedhosseiinii:
`Early return` was used to keep readability


taylorotwell:
If the code is working it's fine.

Issue #407, opened 2022-11-09
Sanctum is returning first and wrong user without middleware and using `Auth::guard('sanctum')`


mikkpokk:
- Sanctum Version: 3.0.1
- Laravel Version: 9.26.1
- PHP Version: 8.1.3
- Database Driver & Version: MySQL 8.0.21

### Description:

Sometimes, there are necessary to lookup user on public route to return extra user related data. For such a cases we have helper `auth('sanctum')->user()`. However, this helper returns the first user from users table, which is not correct. This user is not even token owner.

### Steps To Reproduce:

* Create public route
* Write callback below:
```php
if (auth('sanctum')->check()) {
     return auth('sanctum')->user();
}

return [];
```
* Visit URL as a guest (acts correctly)
* Visit URL with a Bearer token (returns always first user from the table)

driesvints:
Heya, thanks for reporting.

We'll need more info and/or code to debug this further. Can you please create a repository with the command below, commit the code that reproduces the issue **as one separate commit** on the main/master branch and share the repository here? Please make sure that you have the [latest version of the Laravel installer](https://github.com/laravel/installer) in order to run this command. Please also make sure you have both Git & [the GitHub CLI tool](https://cli.github.com) properly set up.

```bash
laravel new bug-report --github="--public"
```

Please do not amend and create a separate commit with your custom changes. After you've posted the repository, we'll try to reproduce the issue.

Thanks!

driesvints:
Are you applying the `auth:sanctum` middleware?

mikkpokk:
No, that is the case. The route should be public, which is why I cannot apply `auth:sanctum` middleware and have to catch user via `Auth::guard`.

The problem isn't that the user is not found. The problem is that $accessToken->tokenable will load first user from the table which is not associated with the token. I assume `auth:sanctum` middleware does relationship eager loading part but guard doesn't.

driesvints:
You can't both have a public and an authed route with Sanctum. The auth middleware resolves the user. Since you're not using it you're running into the message you indicated.

mikkpokk:
Let's imagine the scenario for a second.

I do have public news route and since Authorization header is passed in I do want to pass in whether the user is the owner of the news post or not.

**You're saying `auth('sanctum')->user()` should not work in that scenario and it is correct that it returns the first user from the database table, who is not the token owner?**

mikkpokk:
[...]

PR #408, opened 2022-11-09
Fixes issue #407

mikkpokk:
Read more: Sanctum is returning first and wrong user without middleware and using Auth::guard('sanctum') #407

### DIFF
--- a/src/Guard.php
+++ b/src/Guard.php
@@ -65,9 +65,9 @@ class Guard
             $model = Sanctum::$personalAccessTokenModel;
 
             $accessToken = $model::findToken($token);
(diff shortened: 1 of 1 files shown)


driesvints:
Please reply with a repo on the issue before we continue with this.

taylorotwell:
Unable to recreate and we would have *many* more reports of this if it were truly behaving this way.

mikkpokk:
Problem has existed for a years, you just haven't noticed 🥲 

https://laracasts.com/discuss/channels/laravel/laravel-sanctum-optional-auth-on-route

PR #410, opened 2022-11-23
Carry through session ENV settings to sanctum middleware

chadhutchins:
We have unique use cases where we do not want http_only and same_site hardcoded here. We need them to be specific values that we have set in our ENV. 

This change will allow it to pull the values from the ENV if they exist, if not, fallback to the original values. Please consider, thanks!


taylorotwell:
This could be a breaking change to existing applications by re-using these environment variables for this purpose.

chadhutchins:
I guess I could make my own middleware class that extends this and implements what I need. That makes more sense. Thanks!

PR #411, opened 2022-12-14
Add `CreateFreshApiToken` middleware to Sanctum

tombanksme:
I've seen/heard Taylor question why Laravel users choose Passport over Sanctum a couple of times online. The primary reason why I use Passport in my projects is that Sanctum lacks a `CreateFreshApiToken` middleware.

I usually work on applications with the `laravel/ui` package and Vue components to make the frontend reactive. Importing Passport and adding the `CreateFreshApiToken` middleware is significantly less effort than adding the frontend logic required for Sanctum to work.

Plus I get an oauth-server for free if I need it in the future!

I've hacked together a frankenstein's monster of Sanctum with Passport's `CreateFreshApiToken` middleware. It's obviously not even close to deployment ready but I thought I'd get some feedback before investing serious time:

Questions:
- Would the Laravel team be interested in merging this change?
- Sanctum uses tokens stored in the database. Would the preferred option be for the `CreateFreshApiToken` middleware to create these database tokens rather than JWTs like Passport does?


jayan-blutui:
Have you looked at the `EnsureFrontendRequestsAreStateful` middleware? You could add it to your `api` middleware group to allow frontend requests.

tombanksme:
I didn't actually realise it was commented out. I'll give it a go tomorrow and see if that works.

I want to avoid having to call /sanctum/csrf-cookie if possible.

taylorotwell:
Yeah, the EnsureFrontendRequestsAreStateful is a better version of what I did with CreateFreshApiToken on things like old Spark, etc.

tombanksme:
Thanks for the suggestions 😄 

I can confirm that this worked for me; no need to request `/sanctum/csrf-cookie`. The issue I'd ran into was that my `session.domain` wasn't set correctly which meant that Sanctum wasn't reading the cookie.

Might be worth making this a little more obvious in the docs; not sure I'd have picked up on this had I not spent last night looking through Sanctum's source code.

PR #412, opened 2022-12-14
Allow overriding of default api token length by static property

onlime:
As follow-up for #390, here's a simpler and more flexible solution to override the default token length of 40. Like this, you could simply place a `$apiTokenLength` static property on your model to set the token length.

### DIFF
--- a/src/HasApiTokens.php
+++ b/src/HasApiTokens.php
@@ -47,7 +47,7 @@ trait HasApiTokens
     {
         $token = $this->tokens()->create([
             'name' => $name,
-            'token' => hash('sha256', $plainTextToken = Str::random(40)),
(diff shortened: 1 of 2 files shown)


driesvints:
Why do you need to change this?

RahulDey12:
we could have a config value that can be changed accordingly

onlime:
> we could have a config value that can be changed accordingly

nope, that was already done in PR #390 and rejected in August by @taylorotwell 

And I think my simpler solution is more flexible as we can keep the default 40 chars for all API tokens and only change it for a simple model if needed.

driesvints:
Feel free to resubmit this with a thorough explanation why you need this 👍

PR #414, opened 2023-01-06
Add middleware group configuration

LoniasGR:
We are trying to build an API-only Laravel with an SPA front-end. When removing the `web` middleware routes we figured out Sanctum expects the `web` middleware to exist and adds the `csrf-cookie` route on this middleware group.

This PR amends that by adding a configuration option to change the middleware group to which the endpoint is added to.

### DIFF
--- a/config/sanctum.php
+++ b/config/sanctum.php
@@ -60,6 +60,7 @@ return [
     */
 
     'middleware' => [
+        'group' => 'web',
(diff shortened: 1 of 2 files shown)


jayan-blutui:
I have the same setup as you and what I ended up doing is setting `routes` to `false` in the `sanctum.php` config file, then I created my own csrf cookie route.

driesvints:
You can try what @jayan-blutui suggests.

Issue #418, opened 2023-01-31
Annotation enhancement


btxtiger:
Hi,
the description in the `config/sanctum.php` about `expiration` is a little misleading.

https://github.com/laravel/sanctum/blob/10bade5a481b79adfda000e1430a795625710d0d/config/sanctum.php#L39

We hat the issue, that some tokens suddenly expired even if the `expires_at` was far in the future, and it was really hard to trace back the problem. This happened, because we considered the field `expiration` as default expiration time, which was set to `'expiration' => Carbon::now()->secondsUntilEndOfDay() / 60`.

So in the end, the description 
> If this value is null, personal access tokens do not expire.

is not correct, because the can expire if they have `expires_at` defined.

Imo this field should be refactored to `tokenMaxAgeSec`, but also correcting the description would be already helpful.

It should be also mentioned, that `tokenMaxAgeSec` has a higher priority than `expires_at`.

driesvints:
You can't use a dynamic value like `Carbon::now` in a config file. It should be a fixed value in minutes. Config should be cached in production with `config:cache`.

btxtiger:
> You can't use a dynamic value like Carbon::now in a config file.

It is actually possible as long as it returns a number. But the issue is fixed from my side, and doesn't matter.

I was complaining about the docs/comment/annotation of the parameter. However I'll create a PR to make it more clear.

PR #420, opened 2023-01-31
Fix wrong attribute annotation (#418)

btxtiger:
Detailed derivation inside the issue text.


taylorotwell:
Feels like something more for the full documentation.

btxtiger:
@taylorotwell Probably at that length yes. But the current short comment ist still wrong.
Explicitly `If this value is null, personal access tokens do not expire.` is wrong.

PR #423, opened 2023-02-09
[3.x] PersonalAccessToken::getTokenable() method

Cellard:
This p-r brings `PersonalAccessToken::getTokenable()` method. This opens way to implement custom `getTokenable`.

For example — caching user with tagged cache, that we may flush on User model events.

```php
class PersonalAccessToken extends \Laravel\Sanctum\PersonalAccessToken
{
    public function getTokenable()
    {
        return \Illuminate\Cache\TaggedCache::tags('sanctum')
            ->remember($this->token, now()->addDay(), fn() => $this->tokenable);
    }
}
```

### DIFF
--- a/src/Guard.php
+++ b/src/Guard.php
@@ -67,11 +67,11 @@ class Guard
             $accessToken = $model::findToken($token);
 
             if (! $this->isValidAccessToken($accessToken) ||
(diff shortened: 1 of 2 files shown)


taylorotwell:

PR #424, opened 2023-02-20
PersonalAccessToken fix Find INT

caseroft:
al realizar el find con un int , mejora un 50% la velocidad de consulta.


ankurk91:
What was the issue with original code? Can you reference an issue?

PR #431, opened 2023-03-11
Add missing config example for sanctum routes.

kesoji:
SanctumServiceProvider has useful config keys but these are not listed in config.php.

https://github.com/laravel/sanctum/blob/4a27cd1fe20dde51b3d4eefa02b2f7dffdb512a3/src/SanctumServiceProvider.php#L82-L91

I included them with default values.

### DIFF
--- a/config/sanctum.php
+++ b/config/sanctum.php
@@ -64,4 +64,18 @@ return [
         'encrypt_cookies' => App\Http\Middleware\EncryptCookies::class,
     ],
 
+    /*
+    |--------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


taylorotwell:
I don't think these should really be customized very often. You are free to add them to your config file if you wish.

kesoji:
Thank you for your feedback. I think so too.
In my case, my laravel configuration is for only api usage and all the paths must be under `/api` beacause of the load balancer's configuration so `prefix` configuration was very helpful.

PR #440, opened 2023-04-28
Add default entry for config.middleware.validate_csrf_token

EK1771:
In [v3.2.4](https://github.com/laravel/sanctum/commit/f5bae6156c760545f368438198327e2609ba7bf1#diff-cd82d3386f54185e67e56acc8c7a1e0c537bbc43aec01f51201701eb9dbaf1a4R53) `config.middleware.validate_csrf_token` was added.

This merge request keeps the `config/sanctum.php` file inline with those changes, otherwise if users are taking advantage of excluding URL's for CSRF validation by declaring them in `App\Http\Middleware\VerifyCsrfToken`, it will not be respected and cause a `CRSF token mismatch.` error.

For users upgrading to v3.2.4, this config value will need to be added manually otherwise they will encounter the same issue as above.

### DIFF
--- a/config/sanctum.php
+++ b/config/sanctum.php
@@ -60,6 +60,7 @@ return [
     */
 
     'middleware' => [
+        'validate_csrf_token' => App\Http\Middleware\VerifyCsrfToken::class,
         'verify_csrf_token' => App\Http\Middleware\VerifyCsrfToken::class,
         'encrypt_cookies' => App\Http\Middleware\EncryptCookies::class,
     ],


taylorotwell:
That won't be used yet. Duplicate middleware are removed.

EK1771:
Thanks, but this doesn't seem to be the case as we had issues with v3.2.4 in production without the proposed fix.

Currently the `middleware` array in [EnsureFrontendRequestsAreStateful.php#L49](https://github.com/laravel/sanctum/blob/3.x/src/Http/Middleware/EnsureFrontendRequestsAreStateful.php#L49) now becomes:

```php
array:5 [▼
  0 => "App\Http\Middleware\EncryptCookies"
  1 => "Illuminate\Cookie\Middleware\AddQueuedCookiesToResponse"
  2 => "Illuminate\Session\Middleware\StartSession"
  3 => "Illuminate\Foundation\Http\Middleware\VerifyCsrfToken" <- Shouldn't exist
  4 => "App\Http\Middleware\VerifyCsrfToken"
]
```

However, with the fix:

```php
array:4 [▼
  0 => "App\Http\Middleware\EncryptCookies"
  1 => "Illuminate\Cookie\Middleware\AddQueuedCookiesToResponse"
  2 => "Illuminate\Session\Middleware\StartSession"
  3 => "App\Http\Middleware\VerifyCsrfToken"
]
```

PR #441, opened 2023-04-28
Correct setting key fallback logic

andreladocruz:
This will correctly use of both keys if a different VerifyCsrfToken class is used in sanctum.php file.

### DIFF
--- a/src/Http/Middleware/EnsureFrontendRequestsAreStateful.php
+++ b/src/Http/Middleware/EnsureFrontendRequestsAreStateful.php
@@ -50,8 +50,7 @@ class EnsureFrontendRequestsAreStateful
             config('sanctum.middleware.encrypt_cookies', \Illuminate\Cookie\Middleware\EncryptCookies::class),
             \Illuminate\Cookie\Middleware\AddQueuedCookiesToResponse::class,
             \Illuminate\Session\Middleware\StartSession::class,
-            config('sanctum.middleware.validate_csrf_token', \Illuminate\Foundation\Http\Middleware\VerifyCsrfToken::class),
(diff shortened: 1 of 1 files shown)


RahulDey12:
⚠️ this could lead to breaking changes

andreladocruz:
Why?

taylorotwell:
You will have to elaborate much more on what you are trying to solve.

andreladocruz:
LOL

Let's explain... a little better...

If anyone wants to have their own version of the VerifyCsrfToken class they will need to add it to the sanctum file.

Right?

The idea of having a different VerifyCsrfToken class is that we don't want to load the \Illuminate\Foundation\Http\Middleware\VerifyCsrfToken::class one.

Right?

Your code always adds the "\Illuminate\Foundation\Http\Middleware\VerifyCsrfToken::class" instead we add both keys (validate_csrf_token and verify_csrf_token) to sanctum.php configuration file.

That's why I chained the fallbacks, allowing only one to be added to the sanctum file.

Right?

andreladocruz:
@taylorotwell,

Just saw the new commit.

But I understand your point.

Maybe the line was too long. LOL 

I always have fun creating PRs knowing they will be rejected by you and your team.

andreladocruz:
@taylorotwell.

Just to notice... your commit will have the same problem if someone like me is using an older version of sanctum and still using the "validate_csrf_token" key.

If you really want to avoid errors, you should use my solution.

Or you will need to warn older users to change from "validate_csrf_token" to "verify_csrf_token" key to avoid loading the default VerifyCsrfToken class at all.

image

PR #443, opened 2023-04-29
Use `str_contains`

foremtehan:


taylorotwell:
The code is not broken. Thanks!

PR #449, opened 2023-06-15
Fix return type declaration for currentAccessToken()

albertoelias123:
This PR fixes the return type declaration in the `currentAccessToken()` method. Previously, it didn't account for the fact that `TransientToken` and `PersonalAccessToken` classes also implement the `HasAbilities` interface, causing an inaccurate return type.

To address this, we've updated the return type using Intersection Types from PHP 8.1. Now, the method correctly indicates that it can return an object that is both an `Illuminate\Database\Eloquent\Model` and implements the `Laravel\Sanctum\Contracts\HasAbilities` interface.

### DIFF
--- a/src/Contracts/HasApiTokens.php
+++ b/src/Contracts/HasApiTokens.php
@@ -31,7 +31,7 @@ interface HasApiTokens
     /**
      * Get the access token currently associated with the user.
      *
-     * @return \Laravel\Sanctum\Contracts\HasAbilities
+     * @return \Laravel\Sanctum\PersonalAccessToken|null
      */
     public function currentAccessToken();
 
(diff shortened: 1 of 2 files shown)


albertoelias123:
I'm trying to improve the IDE's autocomplete feature by specifying a more specific return type for the `currentAccessToken()` method. After reviewing the code, I noticed that the method returns an object of type `PersonalAccessToken`, which extends `Model`.

My goal in specifying the return type is to enable the IDE to provide advanced autocomplete suggestions, such as displaying available methods for `PersonalAccessToken` objects. This would allow, for example, the ability to quickly see a `typehint` for the `delete()` method directly in the IDE, without having to refer to the documentation.

This enhancement would make the development process more efficient, as we would have quick and direct access to the specific functionalities of `PersonalAccessToken`, like token revocation, as documented in the "Revoking Tokens" section (https://laravel.com/docs/10.x/sanctum#revoking-tokens) of the Laravel Sanctum documentation.

albertoelias123:
@taylorotwell Would it be more appropriate to leave only the HasAbilities interface?

PR #452, opened 2023-07-13
A comment was written so as not to show an error saying that the property was not found

akbarali1:
When I use phpstorm it can't find the accessToken. That's why these comments are needed.

**Before**
Screenshot 2023-07-13 at 15 44 48

**After**
Screenshot 2023-07-13 at 15 44 19

Screen Recording 2023-07-13 at 15 47 32

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 composer.lock
 /phpunit.xml
 .phpunit.result.cache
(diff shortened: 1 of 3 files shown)


driesvints:
Please resubmit this without any code style changes and without the gitignore change. Thanks

PR #455, opened 2023-07-24
Extract token generation into `generateToken` function

faustbrian:
This is a backward-compatible change that extracts the generation of personal access tokens into a separate function to allow for easier overriding. In most cases, overriding the `createToken` function is excessive because the only part of the creation process worth modifying is the token's generation.

The newly introduced `generateToken` function generates and returns the hashed and plain text token. This approach allows custom algorithms to be implemented without having to replace the entire `createToken` function every time you want to do this. My specific use-case for this change would be to generate PATs that are prefixed with the model type, e.g., `Str::snake(class_basename(static::class))`.

```php
<?php

class User extends Model implements HasApiTokensContract
{
    use HasApiTokens;

    protected function generateToken(string $name, array $abilities, ?\DateTimeInterface $expiresAt)
    {
        $token = hash('sha256', $plainTextToken = TypeId::fromPrefix('user'));

        return [$token, $plainTextToken];
    }
}
```

### DIFF
--- a/src/HasApiTokens.php
+++ b/src/HasApiTokens.php
@@ -45,9 +45,11 @@ trait HasApiTokens
      */
     public function createToken(string $name, array $abilities = ['*'], DateTimeInterface $expiresAt = null)
     {
+        [$token, $plainTextToken] = $this->generateToken($name, $abilities, $expiresAt);
+
         $token = $this->tokens()->create([
             'name' => $name,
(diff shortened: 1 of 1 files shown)


taylorotwell:
Why are a lot of unused arguments sent to the method? Can you resubmit with proper arguments and docblock?

faustbrian:
They were included because they might be relevant to how the token would be generated.

PR #459, opened 2023-08-07
Make tokens identifiable with prefix and checksum

marzvrover:
This PR adds prefixes and CRC32 checksums to tokens.

This makes generated tokens more identifiable to increase security posture by enabling the use of secret scanning technologies such as [GitHub's secret scanning](https://docs.github.com/en/code-security/secret-scanning/about-secret-scanning). With custom prefixes this allows developers onboard secrets they mint to to the [GitHub secret scanning partner program](https://docs.github.com/en/code-security/secret-scanning/secret-scanning-partner-program).

https://github.com/laravel/framework/discussions/47977

Disclosure: I work for GitHub and I am the main engineer involved with the secret scanning partner program.


tests/GuardTest.php
```diff
@@ -542,7 +542,7 @@ protected function getPackageProviders($app)
         return [SanctumServiceProvider::class];
     }
 
-    public function invalidTokenDataProvider(): array
+    public static function invalidTokenDataProvider(): array
```
marzvrover, line 545: This fixes a phpunit warning, but I'm happy to separate it from the PR if it is too off topic for the PR.

taylorotwell:
Thanks 👍

src/HasApiTokens.php
```diff
@@ -45,9 +45,16 @@ public function tokenCan(string $ability)
      */
     public function createToken(string $name, array $abilities = ['*'], DateTimeInterface $expiresAt = null)
     {
+        $plainTextToken = sprintf(
+            '%s%s%s',
+            config('sanctum.token_prefix', ''),
```
mkarnicki, line 50: This will never be empty, because the vendor's sanctum.php defines a `laravel_sanctum_` default value, regardless if the env var is defined or not.

If I understand correctly, this has been included in a minor version release 3.2.6. We use Laravel Sanctum version 3.2. We accidentally realized the tokens now have a "laravel_sanctum_" prefix unless the env var is redefined with empty content, which is unexpected.

How is this not a breaking change?

Thanks for any feedback you may have.

stefro: Thanks @driesvints!

mkarnicki: Thank you! I'm afk, but based on the change in the code the unexpected change in behavior should be gone.

Issue #142, opened 2020-05-15
"logoutOtherDevices" doesn't work with Laravel Sanctum


IlCallo:
- Sanctum Version: 2.3.1
- Laravel Version: 7.11.0
- PHP Version: 7.4.4
- Database Driver & Version: MariaDB 10.4.12

### Description:
`logoutOtherDevices` doesn't work when used in conjunction with Laravel Sanctum.
I'm trying to use it to logout a specific user from all its sessions, yet sessions aren't invalidated.

### Steps To Reproduce:
```php
Auth::guard('web')->setUser(User::find(1))->logoutOtherDevices(Str::random(12));
```

Login with an user, use tinker to run previous command, notice it produces no effect on currently active sessions.

Ragash:
https://laravel.com/docs/7.x/authentication#invalidating-sessions-on-other-devices

have you uncommented the middleware?

IlCallo:
Uncommenting the middeware actually breaks the logout.
Sanctum uses its own Middleware in which "AutenticatesSession" isn't present, and that's probably the reason it fails.
Maybe it could be added in its dynamic Middleware stack

Ragash:
replying in IT for semplicity: 
se leggi le docs, se vuoi utilizzare quel metodo, sei costretto ad abilitare AutenticatesSession middleware.
Sono dipendenti, non c'è altro modo purtroppo.
E si, se lasci Sanctum allo stato brado, assume che il tuo logout sia effettuato sulla sua guard (dove il metodo non esiste) e quindi muore.
la soluzione è semplice:
nel tuo login controller specifica molto banalmente:
```
     /**
     * Get the guard to be used during authentication.
     *
     * @return \Illuminate\Contracts\Auth\StatefulGuard
     */
    protected function guard()
    {
        return \Auth::guard('web');
    }
```
Ovviamente assumendo che tu stia utilizzando il tratto con cui lo scaffolding di laravel viene servito.
prova a fare backtracing dell'errore, cosi capisci bene dove si trova il culprit, le mie assunzione si basano principalmente su esperienza personale (anch'io c ho sbattuto il muso)

IlCallo:
It's simpler for us to discuss in IT, but this won't help next ones hitting this problem, so I'll continue in EN.

I know the feature is tied with AuthenticatesSession, I opened this issue explicitly to make everyone aware this is a missing piece into Sanctum integration with Laravel.

[...]

PR #461, opened 2023-08-10
Ensure device has not been logged out

patrickomeara:
This adds a middleware to check that the password hash in session is the same as the current users password, and logs the user out if not.

This fixes a security issue where an attacker can keep sending requests to an API using the sanctum auth after the password has been updated.

Scenario:
If a user is logged in in two different browsers, they change their password in one, the web session is invalidated in the other, but requests can still be sent via the sanctum cookie auth to the API.

refs #142

### DIFF
--- a/src/Http/Middleware/EnsureDeviceHasNotBeenLoggedOut.php
+++ b/src/Http/Middleware/EnsureDeviceHasNotBeenLoggedOut.php
@@ -21,11 +21,10 @@ class EnsureDeviceHasNotBeenLoggedOut
      */
     public function handle(Request $request, Closure $next): Response
     {
-        if (! $request->hasSession() || ! $request->user()) {
-            return $next($request);
-        }
-
-        if ($request->session()->get('password_hash_'.$this->auth->getDefaultDriver()) !== $request->user()->getAuthPassword()) {
+        if ($request->hasSession()
+            && $request->user()
+            && $request->session()->has($key = 'password_hash_'.$this->auth->getDefaultDriver())
+            && $request->session()->get($key) !== $request->user()->getAuthPassword()) {
             $this->logout($request);
 
             throw new AuthenticationException('Unauthenticated.', [$this->auth->getDefaultDriver()]);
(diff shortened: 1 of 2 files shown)


taylorotwell:
Can you not just use the already existing `AuthenticateSession` middleware?

https://laravel.com/docs/10.x/authentication#invalidating-sessions-on-other-devices

taylorotwell:
Can you not just use the already existing `AuthenticateSession` middleware?

https://laravel.com/docs/10.x/authentication#invalidating-sessions-on-other-devices

patrickomeara:
Hi @taylorotwell 

From what I can see, adding `auth.session` after `auth:sanctum` uses the `RequestGuard` which doesn't have a `viaRemember` method. 

```
Method Illuminate\Auth\RequestGuard::viaRemember does not exist.
```

There is also a fatal call to logoutCurrentDevice

```
Method Illuminate\Auth\RequestGuard::logoutCurrentDevice does not exist.
```

`EnsureDeviceHasNotBeenLoggedOut` is a stripped down version of `AuthenticateSession` that handles only the passsword hash check and logging out if it doesn't match. It rightfully doesn't handle the remember me nor the storing of the password hash.

As it's quite difficult to test for this I have set up a [test repo](https://github.com/patrickomeara/sanctum-logout-issue) that has both the `auth.session` middleware branch and the new middleware branch. There are setup instructions in the README there.

Related issues
https://github.com/laravel/framework/pull/40988
https://stackoverflow.com/questions/65623617/logoutotherdevices-isnt-working-with-laravel-sanctum/70469518#70469518
https://github.com/laravel/sanctum/issues/142

taylorotwell:
@patrickomeara using this middleware - how does the password hash get stored in the session in the first place? Is it stored by `AuthenticateSession` during the original "web" request?

patrickomeara:
Yes the original web request stores the password hash in the session. This is the existing mechanism that logs out other sessions on password change.

This newly added Sanctum-only middleware will log out the sanctum cookie sessions as well.

I've added a check so if the original request isn't using `AuthenticateSession` the middleware continues.

patrickomeara:
I've updated the [test repo](https://github.com/patrickomeara/sanctum-logout-issue/commit/865ea3fe2fcdcf7cf0366c0e0753c504219ee62a) to use this branch of sanctum.

patrickomeara:
I'm converting this to draft as I've found a condition where the guard changes from web to sanctum when changing the password via a sanctum request and logs the current session out.

patrickomeara:
Closing, as https://github.com/laravel/framework/pull/48056 is a better approach to this problem.

PR #467, opened 2023-08-22
[3.x] Ensure device has not been logged out

crynobone:
Continue from https://github.com/laravel/sanctum/pull/461


src/Http/Middleware/EnsureDeviceHasNotBeenLoggedOut.php
```diff
@@ -0,0 +1,94 @@
...
+     * Handle an incoming request.
+     *
+     * @param  \Closure(\Illuminate\Http\Request): (\Symfony\Component\HttpFoundation\Response)  $next
+     *
+     * @throws \Illuminate\Auth\AuthenticationException
+     */
+    public function handle(Request $request, Closure $next): Response
+    {
+        if ($request->hasSession()) {
+            if ($request->session()->has($key = 'password_hash_'.$this->auth->getDefaultDriver())) {
```
crynobone, line 43: From my test `$this->auth->getDefaultDriver()` always returns `sanctum`, using this logic we can loop `sanctum.guard` config and only logged out if the `$guard` in an instance of `SessionGuard`.

patrickomeara: From the [test repo](https://github.com/patrickomeara/sanctum-logout-issue) I have setup, `$this->auth->getDefaultDriver()` returns `web` on a sanctum request, but will use `sanctum` after the request has processed, as the request changes the default guard via `shouldUse()`.

Once we have added the `tap()` to store the password hash after the request as mentioned in my other comment, we now have two different guards being used. `web` and `sanctum`.

Ray screenshot:
image

We can get around this by caching the password hash key.

```php

    protected function passwordHashKey()
    {
        return $this->passwordHashKey ??= 'password_hash_'.$this->auth->getDefaultDriver();
    }
```

patrickomeara (commented):
Hi @crynobone 

This solution will log out the current sanctum session, when the password is changed via a sanctum guarded request.

See below recording.

https://github.com/laravel/sanctum/assets/571773/4e56611b-8bee-4763-9026-3206e3241aba

There are two reasons. 
* The password hash isn't saved after it is changed like it is when using `AuthenticateSession`
* Sanctum requests change default guard from `web` to `sanctum` after the first check is done.

I have created a new branch on my test repo that uses this pr branch https://github.com/patrickomeara/sanctum-logout-issue/tree/crynobone-solution

Details below.

src/Http/Middleware/EnsureDeviceHasNotBeenLoggedOut.php
crynobone: Changed the middleware to only check for auth guards from `sanctum.guard` on only logged out or store password hash on `SessionGuard`.

src/Http/Middleware/EnsureFrontendRequestsAreStateful.php
```diff
@@ -52,6 +52,7 @@ protected function frontendMiddleware()
             \Illuminate\Session\Middleware\StartSession::class,
             config('sanctum.middleware.validate_csrf_token'),
             config('sanctum.middleware.verify_csrf_token', \Illuminate\Foundation\Http\Middleware\VerifyCsrfToken::class),
+            config('sanctum.middleware.ensure_not_logged_out', EnsureDeviceHasNotBeenLoggedOut::class),
```
crynobone, line 55: I feel it might be best to use the same conversion as Laravel Jetstream:

https://github.com/laravel/jetstream/blob/a57287944cd0168cd6e8f9237ef4965711748515/routes/inertia.php#L27-L29

tests/Controller/FrontendRequestsAreStatefulTest.php
```diff
@@ -0,0 +1,113 @@
...
+
+            return $request->user()->email;
+        })->middleware($apiMiddleware);
+
+        $router->get('web/user', function (Request $request) {
+            abort_if(is_null($request->user()), 401);
+
+            return $request->user()->email;
+        })->middleware($webMiddleware);
+    }
```
patrickomeara, line 52: We need to assert that the user is still logged in after a sanctum request changes the password.

```suggestion

        $router->post('/sanctum/api/password', function (Request $request) {
            abort_if(is_null($request->user()), 401);

            $request->user()->update(['password' => bcrypt('laravel')]);

            return $request->user()->email;
        })->middleware($apiMiddleware);
    }
    
    public function test_middleware_keeps_session_logged_in_when_sanctum_request_changes_password()
    {
        $user = UserFactory::new()->create();

        $this->actingAs($user)
            ->getJson('/web/user', [
                'origin' => config('app.url'),
            ])
            ->assertOk()
            ->assertSee($user->email);

        $this->getJson('/sanctum/api/user', [
            'origin' => config('app.url'),
        ])
            ->assertOk()
            ->assertSee($user->email);

        $this->postJson('/sanctum/api/password', [
            'origin' => config('app.url'),
        ])
            ->assertOk()
            ->assertSee($user->email);

        $this->getJson('/sanctum/api/user', [
            'origin' => config('app.url'),
        ])
            ->assertOk()
            ->assertSee($user->email);
    }
```

patrickomeara (commented):
Hi @crynobone 

Does this recent solution deal with changing the password in a sanctum request and leaving the sanctum session logged in?

src/Http/Middleware/AuthenticateSession.php
```diff
@@ -0,0 +1,82 @@
...
+
+                $request->session()->flush();
+
+                throw new AuthenticationException('Unauthenticated.', [...$shouldLoggedOut->keys()->all(), 'sanctum']);
+            }
+
+            $this->storePasswordHashInSession($request, $guards->keys()->first());
+        }
+
+        return $next($request);
```
patrickomeara, line 62: The logging out issue in the failed test, is why the tap is necessary here. Just like `AuthenticateSession`.

The new password needs to be stored after it has been changed.

crynobone: Fixed

crynobone:
> Does this recent solution deal with changing the password in a sanctum request and leaving the sanctum session logged in?

Fixed

src/Http/Middleware/AuthenticateSession.php
```diff
@@ -0,0 +1,88 @@
...
+
+                $request->session()->flush();
+
+                throw new AuthenticationException('Unauthenticated.', [...$shouldLoggedOut->keys()->all(), 'sanctum']);
+            }
+        }
+
+        return tap($next($request), function () use ($request, $guards) {
+            if (! is_null($request->user())) {
+                $this->storePasswordHashInSession($request, $guards->keys()->first());
```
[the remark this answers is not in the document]
crynobone: I feel this would be best to be handled when an issue arises.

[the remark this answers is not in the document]
patrickomeara: Cool. Nice collaborating with you 👍🏻

src/Http/Middleware/AuthenticateSession.php
```diff
@@ -0,0 +1,88 @@
...
+    {
+        if (! $request->hasSession() || ! $request->user()) {
+            return $next($request);
+        }
+
+        $guards = Collection::make(Arr::wrap(config('sanctum.guard')))
+            ->mapWithKeys(fn ($guard) => [$guard => $this->auth->guard($guard)])
+            ->filter(fn ($guard) => $guard instanceof SessionGuard);
+
+        if ($request->hasSession()) {
```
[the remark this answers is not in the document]
crynobone: The response logic is different here.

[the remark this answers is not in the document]
patrickomeara: `$request->hasSession()` will always be true, otherwise it would return on line 44. It can be removed.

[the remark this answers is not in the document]
crynobone: This would caused an issue if `$request->user()` is not `null` while `$request->session()` is `null`.

CleanShot 2023-08-23 at 18 58 06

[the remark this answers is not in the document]
patrickomeara: I think we may be talking about different things 😄 

If `$request->hasSession()` is false, the condition on line 43 is true and the method will return on line 44. Line 46+ can only be reached when `$request->hasSession()` is true.

image

crynobone, line 51:
```suggestion
        if ($request->user()) {
```

PR #475, opened 2023-10-04
Skip ensure stateful middleware if token present

antja0:
Previous pull request #473 

If [SPA Authentication](https://laravel.com/docs/10.x/sanctum#spa-authentication) is used, documentation suggests [enabling EnsureFrontendRequestsAreStateful middleware](https://laravel.com/docs/10.x/sanctum#sanctum-middleware) for api and populating first party endpoints in SANCTUM_STATEFUL_DOMAINS env variable.

This means token authentication from these domains does not work.

What if you want to use both [SPA Authentication](https://laravel.com/docs/10.x/sanctum#spa-authentication) and [API Token Authentication](https://laravel.com/docs/10.x/sanctum#api-token-authentication) from the same domain?

Consider following scenario:

- I want to use SPA Authentication but at the same time provide an interactive API documentation where users can try API endpoints with token
- javascript sends the request with token
- request fails because of invalid state

One possible solution is to host this documentation on another domain.

Is there a reason to check for csrf for API endpoints if _valid_ bearer token is provided?

Rare scenario, but something to consider.

### DIFF
--- a/src/Http/Middleware/EnsureFrontendRequestsAreStateful.php
+++ b/src/Http/Middleware/EnsureFrontendRequestsAreStateful.php
@@ -4,6 +4,7 @@ namespace Laravel\Sanctum\Http\Middleware;
 
 use Illuminate\Routing\Pipeline;
 use Illuminate\Support\Collection;
+use Illuminate\Support\Facades\Auth;
 use Illuminate\Support\Str;
 
 class EnsureFrontendRequestsAreStateful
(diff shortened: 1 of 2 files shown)


taylorotwell:
I think Sanctum is fine. To be honest, I don't fully even follow your issue.

antja0:
@taylorotwell 
> I don't fully even follow your issue.

If SANCTUM_STATEFUL_DOMAINS contains test.com you cannot use API tokens in test.com

If you'd want both Session and API token authentication for the same domain, it is not possible.

If client includes an Authorization header / API token, it's likely they intent to Authorize with the Authorization header / API token.

Requiring session in this case I feel is a bit odd, since if attacker already has valid token then it's game lost anyway.

Of course, if you decide that different forms of authentication should be never mixed this way and every domain should only use one type of authentication, that's a completely valid point too. And possibly more robust philosophy too.

Issue #482, opened 2023-12-10
Guest middleware not working properly with Sanctum + Inertia (+ Vue)


GTCrais:
### Sanctum Version

3.3.2

### Laravel Version

10.35.0

### PHP Version

8.2.0

### Database Driver & Version

MySQL 8.0.26 for Windows

### Description

**Platform:** Windows 11, serving apps using Laragon

When using Sanctum with Inertia and Vue 3 (not sure how relevant Vue here is), after logging in, if we manually type in a URL protected by `guest:sanctum` (or just `guest`), we are able to access that route.

### Steps To Reproduce

1. Run this repo locally - https://github.com/GTCrais/sanctum-test (steps to run it in the `readme`)
2. In navbar, click Login
3. Login with provided credentials
4. You will be redirected to Account
5. When you are on the Account page, hit F5 a couple times -- you will remain on the Account page (this will be relevant in a moment)
6. In the navbar click Home, then Login -- you're redirected to the Account page, as you should be
7. In the URL bar, manually visit the `/login` route -- you will actually land on the login page, even though it's protected by `guest` middleware
8. Here's where it gets interesting -- click on the Account again. Now refresh the page with F5. You're redirected to the Login page; the app thinks you're not logged in
9. Click on the Account in the navbar **again** -- you're able to access it. F5 refresh, back to Login again.

This behavior doesn't go away until you re-run `npm run dev`. Once you do that, you're able to refresh pages after you've logged in, but as soon as you go back to the login page by manually accessing it via URL, the issue re-appears.

When using "regular" Laravel auth mechanism instead of Sanctum, i.e. wrapping my routes in `web` middleware group instead of `api` middleware group (which contains Sanctum's middleware), there is no issue.

### Additional info
Additionally, something funny is going on with cookies when manually visiting routes by typing them in the URL bar.

**Example:**
1. Log out and click Home (basically the initial state of the app)
2. Open Inspector in your browser, Application, Cookies, and manually delete all cookies
3. Click in the URL bar and press enter, effectively reloading the home page
4. Cookies don't get re-set
5. Refreshing the app using F5 doesn't re-set cookies
6. Cookies only get reset by clicking on one of the links in the Navbar, which makes an Inertia request as opposed to a "regular" request

Again, when using `web` middleware group (which contains Inertia middleware, but not the Sanctum one) instead of `api` middleware group (which also contains Inertia middleware, in addition to the Sanctum one), there is no issue.

### Aditional research
[...]

PR #483, opened 2023-12-10
Fix for issue #482

GTCrais:
Ensures the domain is set correctly when determining if the given request is from the first-party application frontend.

https://github.com/laravel/sanctum/issues/482


taylorotwell:
Host and Referer/Origin seems like pretty difference concepts? Not sure we need to make this change.

GTCrais:
Ok, but at least consider reviewing the issue it addresses -- https://github.com/laravel/sanctum/issues/482

PR #485, opened 2023-12-10
feat: expiration by seconds

aditbisa:
The only use case that I can think for this feature (other than for my interview code challenge) is:

- REST API app:
  - That means stateless.
  - Auth using bearer token only.
- And backend need to provide a broadcast endpoint for Websocket / EventSource (SSE):
  - Web API for Websocket / EventSource (SSE) cannot send custom headers.
  - As it is stateless, cookies & session is out of options.
  - The only option is token send as URL query.
- And token for Websocket / EventSource (SSE):
  - One time only.
  - **Short lived (< 10 seconds)**

### DIFF
--- a/config/sanctum.php
+++ b/config/sanctum.php
@@ -44,6 +44,8 @@ return [
     | considered expired. This will override any values set in the token's
     | "expires_at" attribute, but first-party sessions are not affected.
     |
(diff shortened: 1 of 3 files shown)


taylorotwell:

PR #492, opened 2024-01-05
Sanctum configuration value missing

ninjaparade:
## Issue

in the `SanctumServiceProvider`, the route prefix for sanctum is used from the config value `('sanctum.prefix')` but that value is not actually in the config file.

## Solution

Users are unable to find the value in the config, although adding it is simple, it would make sense to provide all configuration values with their defaults in the published config file for ease of use and better DX.

this PR adds that value so customization can be done with less confusion.

https://github.com/laravel/sanctum/blob/02ceac431eb99e7cf1d863e3fc4d3e72750cc720/src/SanctumServiceProvider.php#L86C43-L86C57

### TODO
- [ ] Add proper cascading comment block in config.


taylorotwell:
I personally don't really think this should be changed, so I didn't expose it in the configuration file.

PR #498, opened 2024-02-10
Matching method to contract for createToken()

gammamatrix:
## Errors

The contract does not include the third parameter `DateTimeInterface $expiresAt`.

Contract:
```php
public function createToken(string $name, array $abilities = ['*']);
```
Trait:
```php
public function createToken(string $name, array $abilities = ['*'], DateTimeInterface $expiresAt = null)
```

From PHPStan:
```
Error: Method Laravel\Sanctum\Contracts\HasApiTokens::createToken() invoked with 3 parameters, 1-2 required.
```

***Should this pull request be made against 3.x instead?***

### DIFF
--- a/src/Contracts/HasApiTokens.php
+++ b/src/Contracts/HasApiTokens.php
@@ -2,6 +2,8 @@
 
 namespace Laravel\Sanctum\Contracts;
 
+use DateTimeInterface;
+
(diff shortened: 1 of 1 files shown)


taylorotwell:
Likely because contract couldn't be updated on a patch release.

PR #501, opened 2024-03-05
Laravel 11.x Compatibility

anilkumarthakur60:
dependencies to be compatible with Laravel 11.x.


crynobone:
Laravel 11 compatibilty is available on `master` branch: https://github.com/laravel/sanctum/tree/master

use `composer require "laravel/sanctum:dev-master"` if you need to test it right away.

PR #509, opened 2024-04-06
Add regenerateToken method to HasApiTokens

FatefulNur:
This PR introduces new `regenerateToken` method to the `HasApiTokens` traits in **Laravel Sanctum**. This method allows users to regenerate a new personal access token while revoking the existing one.

### Changes:

- Added `regenerateToken` method to the `HasApiTokens` traits and contract.
- Implemented functionality to delete existing tokens and generate a new one with specified attributes.

### Use-Case

For scenarios requiring frequent rotation of access tokens to mitigate security risks, the addition of a `regenerateToken` method to Laravel Sanctum can greatly simplify token management without manually creating it into our own model.

### Testing:

A unit test have been added to ensure the proper functionality of the `regenerateToken` method.

### DIFF
--- a/src/Contracts/HasApiTokens.php
+++ b/src/Contracts/HasApiTokens.php
@@ -32,6 +32,16 @@ interface HasApiTokens
     public function createToken(string $name, array $abilities = ['*'], DateTimeInterface $expiresAt = null);
 
     /**
+     * Regenerate new personal access token for the user.
+     *
+     * @param  string  $name
+     * @param  array  $abilities
+     * @param  \DateTimeInterface|null  $expiresAt
+     * @return \Laravel\Sanctum\NewAccessToken
+     */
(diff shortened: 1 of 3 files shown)


RahulDey12:
why it's deleting all the token it should delete the transit token instead of all.

FatefulNur:
@RahulDey12 thanks for your advice. Could you tell what's the regard to delete only the transient token?

RahulDey12:
It mean that we are logging out.

FatefulNur:
@RahulDey12 yes said absolute. But `TransientToken` Isn't store into the database. So deleting it might meaning less. Should I close this PR then? Or You have any other opinion about this?

FatefulNur:
May be the idea wasn't relevant with api tokens.

PR #508, opened 2024-04-06
fix user not being able to logout

GigaGiorgadze:
issue description: when user logins as user A  and then logs out and tries to login as user B they can not send request to route protected by auth:sanctum middleware

how to reproduce: i prepared two repositories ([back](https://github.com/GigaGiorgadze/sanctum-logout-bug-back), [front](https://github.com/GigaGiorgadze/sanctum-logout-bug-front)) for reproducing, all you have to do is install backend and frontend repositories, following install guideline in the readme and then press buttons on frontend in following order: login gmail. fetch me, logout, login redberry, fetch me. this second "fetch me" should return 401 without this fix applied. instead of going  and changing vendor i crated custom middleware which implements same fix so you can just replace  'authenticate_session' in config/sanctum.php with this `'authenticate_session' => App\Http\Middleware\AuthenticateSanctumRequest::class,` and issue will be fixed

video proof without fix: 

https://github.com/laravel/sanctum/assets/75663118/54282905-585d-4af7-8f16-e3e3cf895b26

video proof with fix: 

https://github.com/laravel/sanctum/assets/75663118/82672379-dd1a-4abb-96fb-67a33e1da876

i believe  this issue was caused because user was grabbed from the request even though user was logged out during that request. using auth guard to grab it from current session seemed to fix it

PR #511, opened 2024-04-08
Fix/unable to logout

GigaGiorgadze:
this is follow up to previous [pull request](https://github.com/laravel/sanctum/pull/508)

# Issue description
`AuthenticateSession` middleware sets password hash in session after request is sent but uses user from request, this results password hash not being removed from the session after loggout.
this does not cause user to be counted as authenticated but in case user tries to log in with different user following request will return 401 because passwords do not match.

# how to replicate
i created two repositories for this, to replicate all u have to do is configure them locally according to readme. [back](https://github.com/GigaGiorgadze/sanctum-logout-bug-back) and [front](https://github.com/GigaGiorgadze/sanctum-logout-bug-front). after installing this you you should go in frontend and click buttons in following order: "login gmail", "fetch me", "logout", "login redberry", "fetch me". this last request from frontend will fail without this fix.  here is video:   

https://github.com/laravel/sanctum/assets/75663118/701122eb-60fe-4241-9f34-651914ed29a3

# why it does not break any existing features
all previous tests are passing and all i changed was place of getting user, meaning i changed getting user from request to getting them from auth() guard which is up to date after log out.

# tests
i added one test case in FrontendRequestsAreStatefulTest which verifies that password_hash_web is removed if user was logged out during that request and is set if user was not logged out during that request.

without fix implemented this line fails: 

```php
$this->getJson('/sanctum/api/logout', [
            'origin' => config('app.url'),
        ])->assertOk()->assertSee('logged out')->assertSessionMissing('password_hash_web');
```

### DIFF
--- a/src/Http/Middleware/AuthenticateSession.php
+++ b/src/Http/Middleware/AuthenticateSession.php
@@ -64,11 +64,23 @@ class AuthenticateSession
         }
 
         return tap($next($request), function () use ($request, $guards) {
-            $guard = $this->getUsersGuard($guards->keys());
+            $guard = $this->getFirstGuardWithUser($guards->keys());
+
             if (! is_null($guard)) {
                 $this->storePasswordHashInSession($request, $guard);
             }
         });
+    }
+
+    /**
+     * Get the first authentication guard that has a user.
+     *
+     * @param  \Illuminate\Support\Collection  $guards
+     * @return string|null
+     */
+    protected function getFirstGuardWithUser(Collection $guards)
+    {
+        return $guards->first(fn ($guard) => $this->auth->guard($guard)->hasUser());
     }
 
     /**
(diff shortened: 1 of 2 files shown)


taylorotwell:
@GigaGiorgadze Why are you using the global `auth` helper if you already have an `AuthFactory` instance injected into the middleware? Please mark as ready for review when you want me to look again.

GigaGiorgadze:
> @GigaGiorgadze Why are you using the global `auth` helper if you already have an `AuthFactory` instance injected into the middleware? Please mark as ready for review when you want me to look again.

@taylorotwell no reason besides the fact that i did not notice :sweat:. i changed to using injected auth factory, so i will reopen, let me know if anything else seems off ^^

src/Http/Middleware/AuthenticateSession.php
```diff
@@ -64,8 +64,10 @@ public function handle(Request $request, Closure $next): Response
         }
 
         return tap($next($request), function () use ($request, $guards) {
-            if (! is_null($request->user())) {
-                $this->storePasswordHashInSession($request, $guards->keys()->first());
+            $guard = $guards->keys()->first();
+
+            if ($this->auth->guard($guard)->hasUser()) {
```
taylorotwell: Sanctum can be configured to attempt to use an array of guards to authenticate users. How do you know the first guard in the list is the one that actually was used and is the one that has the user?

GigaGiorgadze: good point :eyes:. i did not give it much thought since it was like that before. i rewrote it to getting first guard that user is authenticated with.

Issue #506, opened 2024-04-05
supportsToken is tied to a specific implementation detail


nicoqh:
https://github.com/laravel/sanctum/blob/41097ec5b2d9d79df3aabefd28c060f7847b47fc/src/Guard.php#L101-L106

Why does the authenticatable object need to come from a class that uses a specific trait? Wouldn't it be better if we simply ensure that the object implements the necessary behavior for tokens to be issued (i.e. adheres to an interface), like is being done in so many other places?

```php
if (! function_exists('class_uses_recursive')) {
    /**
     * Returns all traits used by a class, its parent classes and trait of their traits.
     *
     * @param  object|string  $class
     * @return array
     */
    function class_uses_recursive($class)
    {
        if (is_object($class)) {
            $class = get_class($class);
        }

        $results = [];

        foreach (array_reverse(class_parents($class) ?: []) + [$class => $class] as $class) {
            $results += trait_uses_recursive($class);
        }

        return array_unique($results);
    }
}

if (! function_exists('trait_uses_recursive')) {
    /**
     * Returns all traits used by a trait and its traits.
     *
     * @param  object|string  $trait
     * @return array
     */
    function trait_uses_recursive($trait)
    {
        $traits = class_uses($trait) ?: [];

        foreach ($traits as $trait) {
            $traits += trait_uses_recursive($trait);
        }

        return $traits;
    }
}

```

driesvints:
Hi there,

Thanks for reporting but it looks like this is a question which can be asked on a support channel. Please only use this issue tracker for reporting bugs with the library itself. If you have a question on how to use functionality provided by this repo you can try one of the following channels:

- [Laracasts Forums](https://laracasts.com/discuss)
- [Laravel.io Forums](https://laravel.io/forum)
- [StackOverflow](https://stackoverflow.com/questions/tagged/laravel)
- [Discord](https://discordapp.com/invite/KxwQuKb)
- [Larachat](https://larachat.co)
- [IRC](https://web.libera.chat/?nick=laravelnewbie&channels=#laravel)

However, this issue will not be locked and everyone is still free to discuss solutions to your problem!

Thanks.

nicoqh:
@driesvints

Not sure how this can be mistaken for a support question.

driesvints:
Hi @nicoqh. You seem to be asking questions about how things work in Sanctum. That's a support question. This issue tracker is only for reporting bugs. There's also no bug template filled in here.

nicoqh:
[...]

PR #515, opened 2024-05-12
Check if $tokenable is valid by checking whether it implements HasApiTokens contract

Ilyes512:
Check if $tokenable is valid by checking whether it implements HasApiTokens contract instead of checking the specific (trait) implementation.

This relates to https://github.com/laravel/sanctum/issues/506. 

If this is acceptable, with or without more changes, it would also require an update to the docs page. This is because you would have to explicitly implement the `HasApiTokens`-interface and not just use the `HasApiTokens`-trait.


taylorotwell:
Could be a breaking change.

Ilyes512:
@taylorotwell It definitely is. So are you saying "maybe" in the next major version? Or are you saying it's not even "worth" it (for now)?

PR #521, opened 2024-06-14
Add missing CSRF cookie route prefix to config file

hsamshod:
Sanctum's `sanctum/csrf-cookie` route prefix (by default 'sanctum') can be changed (`src/SanctumServiceProvider.php:72`), but this config is not present in config.php.

This PR adds this setting to config.php, so developers can find how to change it without Googling 😄


taylorotwell:
You can define this in your own environment file. I don't think it needs to be changed often enough to clutter the configuration file with it.

PR #522, opened 2024-06-15
Adding data column to personalAccessToken Table

Mohamadalani963:
This pull request introduces a new data column to the personal_access_tokens table in Laravel Sanctum. The data column is intended to store additional JSON-encoded information related to personal access tokens.
Changes

    Migration: Added data column in 2019_12_14_000001_create_personal_access_tokens_table.php.
    Model: Updated PersonalAccessToken.php to include data in $fillable and cast it to JSON.
    Token Creation: Modified HasApiTokens.php and Contracts/HasApiTokens.php to accept data in createToken method.
    Testing: Updated HasApiTokensTest.php to include tests for data attribute.
    Factory: Added data attribute in PersonalAccessTokenFactory.php.

Impact

This change allows developers to store additional metadata with personal access tokens, enhancing the flexibility and functionality of token management in Laravel Sanctum.

### DIFF
--- a/database/migrations/2019_12_14_000001_create_personal_access_tokens_table.php
+++ b/database/migrations/2019_12_14_000001_create_personal_access_tokens_table.php
@@ -17,6 +17,7 @@ return new class extends Migration
             $table->string('name');
             $table->string('token', 64)->unique();
             $table->text('abilities')->nullable();
+            $table->text('data')->nullable();
             $table->timestamp('last_used_at')->nullable();
             $table->timestamp('expires_at')->nullable();
             $table->timestamps();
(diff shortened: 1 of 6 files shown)


database/migrations/2019_12_14_000001_create_personal_access_tokens_table.php
```diff
@@ -17,6 +17,7 @@ public function up(): void
             $table->string('name');
             $table->string('token', 64)->unique();
             $table->text('abilities')->nullable();
+            $table->text('data')->nullable();
```
hsamshod, line 20: why not to create as ->json field here?

Mohamadalani963: @hsamshod just following the same as abilities column convention but yes it would be better to make it as json column

driesvints:
We can't do this on a patch release unfortunately as the new column will not be found on existing apps.

PR #523, opened 2024-06-16
feat: add config option to configure remaining middlware

bedus-creation:
Thank you for the excellent package.

This pull request introduces a feature to customize middleware configuration for Illuminate\Session\Middleware\StartSession::class and Illuminate\Foundation\Http\Middleware\ValidateCsrfToken::class.

I understand that while this middleware can be configured using [Laravel's documentation](https://laravel.com/docs/11.x/middleware#laravels-default-middleware-groups), it appears to be only possible for the web group and not for the api group, as these middleware directly applies from `EnsureFrontendRequestsAreStateful`,


taylorotwell:

PR #532, opened 2024-08-07
fix load tokenable relation

### LINKED ISSUE #531: Issue with Auth ID Returning Incorrect User in Route
### Sanctum Version

3.3.3

### Laravel Version

9.52.16

### PHP Version

8.1.0

### Database Driver & Version

_No response_

### Description

**Route Definition**
I have a route defined as follows:

```php
Route::get('/{slug}', [ProductController::class, 'show'])->name('show');
```
This route does not have the auth middleware applied, but an authentication token is sent.

**Controller**
The controller method is as follows:

```php
public function show(int|string $slug, Request $request): JsonResponse
{
    $model = $this->repository
        ->where('slug', $slug)
        ->with([
            'log',
        ])
        ->firstOrFail();
    // ...
}
```

**Relation in Model**
The log relation is defined in the model as follows:

```php
public function log(): MorphOne
{
    return $this->morphOne(Enrollment::class, 'entity')
        ->where('user_id', Auth::id());
}
```

I am using Auth::id() to get the current authenticated user's data. However, Auth::id() is not working correctly and always returns the user with ID 1 from the users table.

Investigation in Guard.php
Upon investigation, I found that in the Guard.php class, it retrieves the tokenable relation like this:

```php
$accessToken->tokenable
```
The query generated is as follows:

```php
array:2 [ // vendor/laravel/sanctum/src/Guard.php:72
  0 => array:3 [
    "query" => "select * from `personal_access_tokens` where `personal_access_tokens`.`id` = ? limit 1"
    "bindings" => array:1 [
      0 => "77217"
    ]
    "time" => 0.45
  ]
  1 => array:3 [
    "query" => "select * from `users` where `users`.`deleted_at` is null limit 1"
    "bindings" => []
    "time" => 0.29
  ]
]
```

##Fix
I made the following change in the PersonalAccessToken model within the findToken method, which resolved the issue:

Instead of this:

```php
$instance = static::find($id)
```

I used:

```php
$instance = static::query()->with('tokenable')->find($id)
```

The resulting query was:

```php
array:2 [ // vendor/laravel/sanctum/src/Guard.php:72
  0 => array:3 [
    "query" => "select * from `personal_access_tokens` where `personal_access_tokens`.`id` = ? limit 1"
    "bindings" => array:1 [
      0 => "77217"
    ]
    "time" => 0.43
  ]
  1 => array:3 [
    "query" => "select * from `users` where `users`.`id` in (55752) and `users`.`deleted_at` is null"
    "bindings" => []
    "time" => 0.48
  ]
]
```

This change fixed the issue, and now Auth::id() returns the correct user ID.

### Steps To Reproduce

1. Create a new Laravel project using `laravel new bug-report`.
2. Add the following route to `routes/web.php`:
    ```php
    Route::get('/{slug}', [ProductController::class, 'show'])->name('show');
    ```
3. Create a `ProductController` with the following method:
    ```php
    public function show(int|string $slug, Request $request): JsonResponse
    {
        $model = $this->repository
            ->where('slug', $slug)
            ->with([
                'log',
            ])
            ->firstOrFail();
        // ...
    }
    ```
4. Define the `log` relationship in the model as follows:
    ```php
    public function log(): MorphOne
    {
        return $this->morphOne(Enrollment::class, 'entity')
            ->where('user_id', Auth::id());
    }
    ```
5. Send a request to the `/show` route with an authentication token.
6. Observe that `Auth::id()` returns the user with ID 1 regardless of the actual authenticated user.

mohammadgh1370:
Fixes an issue with user authentication where the wrong user ID is returned.

Fixes #531

### DIFF
--- a/src/PersonalAccessToken.php
+++ b/src/PersonalAccessToken.php
@@ -63,7 +63,7 @@ class PersonalAccessToken extends Model implements HasAbilities
 
         [$id, $token] = explode('|', $token, 2);
 
-        if ($instance = static::find($id)) {
+        if ($instance = static::query()->with('tokenable')->find($id)) {
             return hash_equals($instance->token, hash('sha256', $token)) ? $instance : null;
         }
     }


taylorotwell:
No tests.

PR #534, opened 2024-08-21
[4.x] Remove `AuthenticateSession::handle` return type

stevebauman:
This PR removes the enforced return type on the `AuthenticateSession::handle` method. This may have accidentally slipped in from a PR or something.

### DIFF
--- a/src/Http/Middleware/AuthenticateSession.php
+++ b/src/Http/Middleware/AuthenticateSession.php
@@ -9,7 +9,6 @@ use Illuminate\Contracts\Auth\Factory as AuthFactory;
 use Illuminate\Http\Request;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Collection;
-use Symfony\Component\HttpFoundation\Response;
 
 class AuthenticateSession
 {
--- a/src/Http/Middleware/AuthenticateSession.php
+++ b/src/Http/Middleware/AuthenticateSession.php
@@ -33,12 +33,13 @@ class AuthenticateSession
     /**
      * Handle an incoming request.
      *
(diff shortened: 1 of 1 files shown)


taylorotwell:
@stevebauman what else could the return type be?

stevebauman:
@taylorotwell Nothing, but the other middleware in this package don't have types:

https://github.com/laravel/sanctum/blob/85684a4eeec272fa0f2a6bd1f20143f3f2306a70/src/Http/Middleware/CheckAbilities.php#L20
https://github.com/laravel/sanctum/blob/85684a4eeec272fa0f2a6bd1f20143f3f2306a70/src/Http/Middleware/CheckForAnyAbility.php#L20
https://github.com/laravel/sanctum/blob/85684a4eeec272fa0f2a6bd1f20143f3f2306a70/src/Http/Middleware/CheckForAnyScope.php#L23
https://github.com/laravel/sanctum/blob/85684a4eeec272fa0f2a6bd1f20143f3f2306a70/src/Http/Middleware/CheckScopes.php#L23
https://github.com/laravel/sanctum/blob/85684a4eeec272fa0f2a6bd1f20143f3f2306a70/src/Http/Middleware/EnsureFrontendRequestsAreStateful.php#L18

Was pushing this just for consistency sake 🙏 

Do you have a rule for when a type-hint/return type should or shouldn't be used so I know for the future?

taylorotwell:
Got it - I think if it's OK I'll just leave it for now since it shouldn't be possible to return anything else. Just wouldn't want to break anything extending it or whatever.

stevebauman:
Ok sounds good! Appreciate the reply 🙏

PR #539, opened 2024-11-01
Chaperone tokenable model

scrummitch:
Since this repository requires laravel v11 lets automatically chaperone the tokenable relationship since its somewhat convoluted to eager load. 

For example:
```php
// Will cause a lazy loading exception if Model::preventLazyLoading() is enabled
$user->tokens->map(fn (PersonalAccessToken $token) {
    dump($token->tokenable);
});
```

### DIFF
--- a/src/HasApiTokens.php
+++ b/src/HasApiTokens.php
@@ -21,7 +21,7 @@ trait HasApiTokens
      */
     public function tokens()
     {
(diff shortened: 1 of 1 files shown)


taylorotwell:
I don't think I really want to introduce this on a patch release.

PR #543, opened 2024-11-13
[4.x] use method instead of array for casts on `PersonalAccessToken`

cosmastech:
I need to be able to add some additional columns to my `personal_access_tokens` table. This makes it possible that I can do the following:

```php
class TeamworksToken extends PersonalAccessToken
{
    protected function casts()
    {
        return parent::casts() + ['is_impersonation' => 'bool'];
    }
}
```

### DIFF
--- a/src/PersonalAccessToken.php
+++ b/src/PersonalAccessToken.php
@@ -10,18 +10,21 @@ class PersonalAccessToken extends Model implements HasAbilities
     /**
      * The attributes that should be cast to native types.
      *
-     * @var array
+     * @return array<string, string>
      */
-    protected $casts = [
-        'abilities' => 'json',
-        'last_used_at' => 'datetime',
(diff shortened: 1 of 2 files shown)


cosmastech:
Taking a look at how https://github.com/laravel/framework/blob/a296e813d041ddff000d856aa2f7ae1cd4d616b7/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php#L196 is setup, for my use case, it wouldn't matter, since a `casts()` method on the child class would get merged in.

Possibly other people would want to manipulate the casts in a different way. 🤷

If you don't want to accept this PR because it's fine as is, I'll make a follow up to just add in the proper typehints on the model.

taylorotwell:
Since it's fine as is we can just do the type-hints PR.

PR #544, opened 2024-11-13
[4.x] Add Generics to `HasApiTokens`

cosmastech:
This adds generics to `HasApiTokens` to aid in static analysis.

This should be beneficial to folks who are overriding the default Sanctum models [as outlined in the docs](https://laravel.com/docs/11.x/sanctum#overriding-default-models), but also should provide IDE typehint && static analysis benefits.

If you are overriding the personal access token model, then you would add this to your tokenable model.

```php
class User extends Authenticatable
{
    /** @use HasApiTokens<TeamworksAccessToken> */
    use HasApiTokens;

    // ...
}
```

Otherwise, the default for the generic is already set.

As a bonus, it also adds the generics for `tokens()` relationship to keep in line with Laravel's MorphMany's generics.

## Outstanding Question
Should the same generics be added to the Contracts/HasApiTokens interface? 🤔

### DIFF
--- a/src/HasApiTokens.php
+++ b/src/HasApiTokens.php
@@ -5,19 +5,22 @@ namespace Laravel\Sanctum;
 use DateTimeInterface;
 use Illuminate\Support\Str;
 
+/**
+ * @template TTokenModel of \Illuminate\Database\Eloquent\Model&\Laravel\Sanctum\Contracts\HasAbilities
+ */
 trait HasApiTokens
 {
     /**
      * The access token the user is using for the current request.
      *
-     * @var \Laravel\Sanctum\Contracts\HasAbilities
+     * @var TTokenModel
      */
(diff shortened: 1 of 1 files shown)


unre4l:
Small question, how does this go with https://github.com/laravel/sanctum/blob/4.x/src/Guard.php#L59 ? `TransientToken` doesnt align with `TTokenModel of \Illuminate\Database\Eloquent\Model` part.

cosmastech:
> Small question, how does this go with https://github.com/laravel/sanctum/blob/4.x/src/Guard.php#L59 ? `TransientToken` doesnt align with `TTokenModel of \Illuminate\Database\Eloquent\Model` part.

Ugh. I think you're right about that 😢 . I'll try a PR to remove the Model part.

PR #547, opened 2024-11-20
Remove default `authenticate_session` middleware

owenconti:
This PR removes the default `authenticate_session` middleware that is set by Sanctum.

We ran into a hard to debug issue in our application where the user would be logged out immediately after impersonating _some_ users via Nova. As it turns out the "some" ended up being based on if the users shared the same password or not. So locally, everything worked fine as all seeded users get the same password. But in production, this is not the case.

I think another potential update here would be to keep the default config as-is, but update the Sanctum docs to make it clear what the `AuthenticateSession` middleware does, and that it is enabled by default. However, it seems like all other areas of the Laravel ecosystem make the "log out other devices" feature opt-in, instead of opt-out.

### DIFF
--- a/config/sanctum.php
+++ b/config/sanctum.php
@@ -75,7 +75,7 @@ return [
     */
 
     'middleware' => [
-        'authenticate_session' => Laravel\Sanctum\Http\Middleware\AuthenticateSession::class,
+        'authenticate_session' => null,
         'encrypt_cookies' => Illuminate\Cookie\Middleware\EncryptCookies::class,
         'validate_csrf_token' => Illuminate\Foundation\Http\Middleware\ValidateCsrfToken::class,
     ],
--- a/tests/Controller/FrontendRequestsAreStatefulTest.php
+++ b/tests/Controller/FrontendRequestsAreStatefulTest.php
@@ -4,6 +4,7 @@ namespace Laravel\Sanctum\Tests\Controller;
 
 use Illuminate\Foundation\Testing\RefreshDatabase;
 use Illuminate\Http\Request;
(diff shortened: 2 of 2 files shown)


taylorotwell:
Hey @owenconti, we probably don't want to remove this from here as it could be a security issue, but instead we should maybe fix it in Nova's impersonation logic.

Do you have an issue if you use Nova's "Stop Impertonating" button? Can you give us a step-by-step of how to recreate?

SRWieZ:
I have a similar issue with Nova.

I had to change `auth:sanctum` to `auth:web` to make it work.

```php
Route::middleware([
    'auth:web',
    config('jetstream.auth_session'),
    EnsureEmailIsVerified::class,
])->group(
// ..
```

I didn't fully understand the problem at the time, but I’m sharing this in case it helps fixing the issue.

owenconti:
@taylorotwell definitely understand not wanting to remove the default value for security reasons, no problem there. 

I think this isn't on Nova, though. Any other package (or custom code) that does impersonation in combination with using Sanctum (and more specifically Sanctum's [`EnsureFrontendRequestsAreStateful` middleware](https://github.com/laravel/sanctum/blob/4.x/src/Http/Middleware/EnsureFrontendRequestsAreStateful.php#L54)) will run into the issue.

Maybe a docs update to the [Sanctum Middleware](https://laravel.com/docs/11.x/sanctum#sanctum-middleware) that explains that the `AuthenticateSession` middleware will be applied by default which results in changes need for impersonation to work correctly would be a safer choice? I'd be happy to draft those doc updates if that's the route you want to go.

taylorotwell:
Hey @owenconti - yeah, I think a doc update could be helpful. Thanks!

PR #553, opened 2024-12-22
Allow Custom Token Length in Sanctum Token Generation

usmanx03:
This update allows the token length in `generateTokenString` to be configurable via the `sanctum.entropy_length` config variable. It defaults to 40 for backward compatibility while enabling developers to define custom lengths for enhanced security.

Custom lengths provide users the flexibility to enhance token security or align with their application's unique requirements, such as stricter compliance standards or integration needs.

The change is fully backwards-compatible and simplifies token customization for varied application needs. All tests passed.


taylorotwell:

usmanx03:
Thank you. I will be putting this up as a separate package.

src/HasApiTokens.php
```diff
@@ -76,12 +76,12 @@ public function createToken(string $name, array $abilities = ['*'], ?DateTimeInt
...
      * @return string
      */
-    public function generateTokenString()
+    public function generateTokenString(): string
     {
         return sprintf(
             '%s%s%s',
             config('sanctum.token_prefix', ''),
-            $tokenEntropy = Str::random(40),
+            $tokenEntropy = Str::random(config('sanctum.entropy_length'), 40),
```
alexleeelkins, line 84: this should be `$tokenEntropy = Str::random(config('sanctum.entropy_length', 40)),`. if `config('sanctum.entropy_length')` is not set, as of now, it will not return a random string.

usmanx03: Yes, but the request is denied, so maybe we should not fix it. Or should I reopen the PR?

PR #564, opened 2025-04-16
Feature to treat same domain requests to be from frontend and make stateful

denjaland:
Our application is running under multiple domain names (multi tenant), each domain name hosting an SPA which communicates to the API endpoints under their own domain name as well.

In order to make sure that Sanctum recognizes these incoming requests as coming from the frontend, today we can only add them to the configuration file, which is cached, and which makes it cumbersome when we want to have that automatcally updated when a user adds a new domain name.

So we were looking into how we could resolve this by overriding the default behaviour of the middleware fromFrontend() method to dynamically check against the tenant host names in our database, but while doing so we actually came to the conclusion that in fact any call that is from the same domain (the referer / origin matches the request host), should be considered as being from the SPA frontend and therefore be made stateful.

Personally, I don't really see a case where this logic would not be applicable, and maybe this could be included into the standard behaviour, but as I'm sure this was thought through when implemented initially, I probably overlook a reason as to why you might not want this behaviour, so I made this PR backwards compatible, by adding a config parameter to enable this behaviour, so it remains disabled by default, and this can be released in a minor.

### DIFF
--- a/config/sanctum.php
+++ b/config/sanctum.php
@@ -12,27 +12,17 @@ return [
     | Requests from the following domains / hosts will receive stateful API
     | authentication cookies. Typically, these should include your local
     | and production domains which access your API via a frontend SPA.
+    | Sanctum::currentRequestHost() instructs Sanctum to include the
+    | request host from the current request in the stateful list.
     |
     */
 
     'stateful' => explode(',', env('SANCTUM_STATEFUL_DOMAINS', sprintf(
         '%s%s',
         'localhost,localhost:3000,127.0.0.1,127.0.0.1:8000,::1',
-        Sanctum::currentApplicationUrlWithPort()
+        Sanctum::currentApplicationUrlWithPort(),
+        // Sanctum::currentRequestHost(),
     ))),
-
-    /*
-    |--------------------------------------------------------------------------
-    | Include Request Host as stateful domain
-    |--------------------------------------------------------------------------
-    |
-    | In most cases frontend SPA implementations call API endpoints on the
-    | same domain name the SPA is hosted from.  This parameter enables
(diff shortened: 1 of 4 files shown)


taylorotwell:
@denjaland I wonder if we could have a `Sanctum::currentRequestHost()` method that you can add to the `stateful` config array... that method would return some placeholder string like CURRENT_REQUEST_HOST which we then find and replace dynamically. Does that make sense? I think that would be cleaner than a new config option.

denjaland:
@taylorotwell, yes that does make sense, and that was actually my first approach as well.  At some point however, I didn't go forward with it because configuration is cached, and I can't include the dynamic host into the stateful list.
I revisited the code, and now I'm using a fixed token I'm replacing at runtime in the `fromFrontend()` check.

I'm not sure whether you have a standard pattern to use such fixed tokens, so please let me know if you prefer this differently.

denjaland:
(I just noticed you actually proposed using a token now LOL - apparently I didn't read beyond your first sentence; I guess I had too many easter eggs ;-) )

siarheipashkevich:
@denjaland please add the documentation for this

denjaland:
> @denjaland please add the documentation for this

Since there is no documentation on the use of Sanctum::currentRequestHost(), I followed example and didn't create a PR for addinv it into the docs.

I noticed however thag @taylorotwell removed the little documentation I added into the config file, so I agree that maybe we should add it to the documentation. 

Shall I create a PR, @taylorotwell, meanwhile documenting the currejtRequestHost helper as well?

denjaland:
@siarheipashkevich I went ahead and created the PR at https://github.com/laravel/docs/pull/10350.

PR #571, opened 2025-07-09
[Refactor] Add indexes to personal_access_tokens

keshav-k3:
This PR adds database indexes to the `expires_at` and `last_used_at` columns in the `personal_access_tokens` table migration.

### Benefit to end users
Improves query performance for token expiration and pruning, especially at scale. Keeps token management fast and efficient as data grows.
### Why it’s safe
The change is backward-compatible and does not modify any existing data or application logic.
_Indexes are a standard optimization_
### Testing
Existing tests for token creation, expiration, and pruning continue to pass.

### DIFF
--- a/database/migrations/2019_12_14_000001_create_personal_access_tokens_table.php
+++ b/database/migrations/2019_12_14_000001_create_personal_access_tokens_table.php
@@ -17,8 +17,8 @@ return new class extends Migration
             $table->text('name');
             $table->string('token', 64)->unique();
             $table->text('abilities')->nullable();
-            $table->timestamp('last_used_at')->nullable();
-            $table->timestamp('expires_at')->nullable();
+            $table->timestamp('last_used_at')->nullable()->index();
+            $table->timestamp('expires_at')->nullable()->index();
             $table->timestamps();
         });
     }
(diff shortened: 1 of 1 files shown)


patrickomeara:
Hopefully this saves someone even a tiny bit of time.

```php
<?php

use Illuminate\Database\Migrations\Migration;
use Illuminate\Database\Schema\Blueprint;
use Illuminate\Support\Facades\Schema;

return new class extends Migration
{
    public function up(): void
    {
        Schema::table('personal_access_tokens', function (Blueprint $table) {
            $table->index('expires_at');
        });
    }
};
```

kylemilloy:
```php
    /**
     * Reverse the migrations.
     *
     * @return void
     */
    public function down()
    {
        Schema::table('personal_access_tokens', function (Blueprint $table) {
            $table->dropIndex(['expires_at']);
        });
    }
```

And for the down.

wsamoht:
Just reviewing changes doing a `composer update` in our app and curious why Taylor removed the index on `last_used_at` 🤔 Maybe he just didn't want it as a default? Obviously we can still add it ourselves.

Issue #577, opened 2026-01-06
Sanctum AuthenticateSession incompatible with Laravel 12.45.0 password hash HMAC change


ams-ryanolson:
## Description

Laravel Framework v12.45.0 (PR laravel/framework#58107) changed how password hashes are stored in sessions - they're now stored as HMACs instead of raw hashes for security. However, Sanctum's `AuthenticateSession` middleware was not updated to match, causing authentication failures.

## The Conflict

**Laravel's `AuthenticateSession`** (updated in framework#58107) now stores the password hash as an HMAC:
```php
// Illuminate\Session\Middleware\AuthenticateSession::storePasswordHashInSession()
$request->session()->put([
    'password_hash_'.$this->auth->getDefaultDriver() => $this->guard()->hashPasswordForCookie($request->user()->getAuthPassword()),
]);
```

**Sanctum's `AuthenticateSession`** still stores and compares the **raw password hash**:
```php
// Laravel\Sanctum\Http\Middleware\AuthenticateSession::storePasswordHashInSession()
$request->session()->put([
    "password_hash_{$guard}" => $this->auth->guard($guard)->user()->getAuthPassword(),
]);
```

## Reproduction

1. Use Laravel 12.45.0+ with Sanctum stateful SPA authentication
2. Enable `$middleware->authenticateSessions()` in bootstrap/app.php
3. Login via web route → Laravel's middleware stores `password_hash_web` as **HMAC**
4. Make API request → Sanctum's middleware reads `password_hash_web`, compares against **raw hash** → mismatch → 401 Unauthenticated

## Expected Behavior

Sanctum's `AuthenticateSession` should use `hashPasswordForCookie()` and include backward compatibility for raw hashes, similar to how Laravel's version was updated.

## Suggested Fix

Update `Laravel\Sanctum\Http\Middleware\AuthenticateSession` to:
1. Store password hashes using `$guard->hashPasswordForCookie()`
2. Add a `validatePasswordHash()` method that tries HMAC first, falls back to raw hash comparison

## Related

- laravel/framework#58107 - Original change
- laravel/framework#58296 - Issue I opened before finding the root cause

## Environment

- Laravel Framework: 12.45.0
- Sanctum: 4.x
- PHP: 8.3

ams-ryanolson:
Submitted a fix: #578

PR #578, opened 2026-01-06
Support HMAC password hash format from Laravel 12.45.0+

ams-ryanolson:
## Summary

Laravel Framework v12.45.0 (PR laravel/framework#58107) changed how password hashes are stored in sessions - they're now stored as HMACs instead of raw hashes for improved security.

This updates Sanctum's `AuthenticateSession` middleware to match the framework's approach.

## Changes

1. **`storePasswordHashInSession()`** - Uses `hashPasswordForCookie()` when available (Laravel 12.45.0+), falls back to raw hash for older versions
2. **`validatePasswordHash()`** - New method that tries HMAC format first, falls back to raw hash comparison for backward compatibility

## Problem

When both `$middleware->authenticateSessions()` and Sanctum stateful auth are enabled:

1. Login via web route → Laravel's `AuthenticateSession` stores `password_hash_web` as **HMAC**
2. API request → Sanctum's `AuthenticateSession` compares against **raw hash** → mismatch → 401

## Backward Compatibility

- Uses `method_exists()` checks so it works with Laravel versions before and after 12.45.0
- Validation tries HMAC first, falls back to raw hash, so existing sessions continue to work

## Related

- laravel/framework#58107 - Original framework change
- #577 - Issue describing this incompatibility

### DIFF
--- a/src/Http/Middleware/AuthenticateSession.php
+++ b/src/Http/Middleware/AuthenticateSession.php
@@ -50,8 +50,11 @@ class AuthenticateSession
         $shouldLogout = $guards->filter(
             fn ($guard, $driver) => $request->session()->has('password_hash_'.$driver)
         )->filter(
-            fn ($guard, $driver) => $request->session()->get('password_hash_'.$driver) !==
-                                    $request->user()->getAuthPassword()
+            fn ($guard, $driver) => ! $this->validatePasswordHash(
+                $guard,
+                $request->user()->getAuthPassword(),
+                $request->session()->get('password_hash_'.$driver)
+            )
         );
 
         if ($shouldLogout->isNotEmpty()) {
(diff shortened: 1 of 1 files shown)


rdehnhardt:
Hello,

This PR addresses the HMAC hash compatibility, which is a great improvement for security.

I wanted to share a related issue and a helpful workaround for users implementing passwordless authentication (such as magic links or OTP), where the user record in the database has a `null` or empty password field.

In a passwordless setup, you might still encounter issues with `AuthenticateSession` when it attempts to validate the session state for an authenticated user who is intentionally passwordless.

To prevent potential errors (like a `TypeError` or unexpected hash comparison failure), users can implement a custom middleware that explicitly bypasses the session validation for authenticated users whose password is `null` or an empty string.

Here is the custom middleware to use instead of the default `Laravel\Sanctum\Http\Middleware\AuthenticateSession`:

```php
<?php

// ...

final class AuthenticateSession extends SanctumAuthenticateSession
{
    /**
     * Handle an incoming request.
     *
     * @param  Request  $request
     * @param  Closure(Request): Response  $next
     * @return Response
     * @throws AuthenticationException
     */
    public function handle(Request $request, Closure $next): Response
    {
        if ($this->shouldSkipAuthentication($request)) {
            return $next($request);
        }

        return parent::handle($request, $next);
    }

    /**
     * Determine if the authentication should be skipped.
     *
     * Skip when the user is authenticated but has no password (passwordless users).
     */
    private function shouldSkipAuthentication(Request $request): bool
    {
        $user = $request->user();

        if ($user === null) {
            return false;
        }

        return $user->getAuthPassword() === null || $user->getAuthPassword() === '';
    }
}
```

This ensures that session validation is correctly skipped when the authenticated user is intentionally passwordless.

src/Http/Middleware/AuthenticateSession.php
```diff
@@ -94,8 +97,33 @@ protected function getFirstGuardWithUser(Collection $guards)
...
+
+    /**
+     * Validate the password hash against the stored value.
+     *
+     * @param  \Illuminate\Auth\SessionGuard  $guard
+     * @param  string  $passwordHash
+     * @param  string  $storedValue
+     * @return bool
+     */
+    protected function validatePasswordHash(SessionGuard $guard, string $passwordHash, string $storedValue): bool
```
patrickomeara, line 117: `$passwordHash` may be null here. 

```
Laravel\Sanctum\Http\Middleware\AuthenticateSession::validatePasswordHash(): Argument #2 ($passwordHash) must be of type string, null given, called in vendor/laravel/sanctum/src/Http/Middleware/AuthenticateSession.php on line 53
```

patrickomeara: Update: My user factory wasn't setting a password so some of my tests were failing with the above error. 

As @rdehnhardt mentioned there are cases where password is null for real world users so perhaps the `$passwordHash` type should be `?string` or have the type removed. Both `SessionGuard::hashPasswordForCookie()` and `hash_equals` do not enforce the `string` type.

@taylorotwell I've opened a PR #581 to address this.

In my case I've just set a password in my factory.

ams-ryanolson:
Why would you not just extend it? Just asking. Setting this to ?string just seems like a bad idea all around.

__________________________
Ryan Olson
Co-founder / CEO
Arctic Media Solutions Inc.

Sent from my mobile

________________________________

PR #581, opened 2026-01-11
Allow null password hash

patrickomeara:
The users password may be null at this point.

This PR fixes a type error introduced in #578 where a user may be logged in using passwordless authentication (such as magic links or OTP)

### DIFF
--- a/src/Http/Middleware/AuthenticateSession.php
+++ b/src/Http/Middleware/AuthenticateSession.php
@@ -110,11 +110,11 @@ class AuthenticateSession
      * Validate the password hash against the stored value.
      *
      * @param  \Illuminate\Auth\SessionGuard  $guard
-     * @param  string  $passwordHash
+     * @param  string|null  $passwordHash
      * @param  string  $storedValue
      * @return bool
      */
-    protected function validatePasswordHash(SessionGuard $guard, string $passwordHash, string $storedValue): bool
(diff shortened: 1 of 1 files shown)


ams-ryanolson:
I would be very concerned about this as part of sanctum. Magic and Passwordless logins are not a standard of security. 

Ideally, someone should extend sanctum if they need this and set it accordingly. 

Alternatively, set a random password for the user in the users table for security but use magic or passwordless logins after.

patrickomeara:
@ams-ryanolson Hey Ryan, your PR was needed and well implemented. While I agree with the points you're making above, I'm just offering this PR up to fix a breaking change it introduced for those that do use those authentication mechanisms. No stress if it's merged or closed.

>Ideally, someone should extend sanctum if they need this and set it accordingly.

Sanctum didn't need to be extended beforehand to handle a null password.

The old argument types in the framework layer conundrum. 

I hope you're having a great weekend.

Tamas-hi:
Thanks for the fix, this was indeed a breaking change for us as well. 

For others coming here: pinning to version 4.2.1 until a new version is released solves this issue.

BnitoBzh:
The fix is not working. 
The `hash_equals` function at the end of `validatePasswordHash` method requires a string for `$passwordHash`.

Must be replaced with : `return hash_equals($passwordHash ?? '', $storedValue);`

PR #586, opened 2026-01-22
[4.x] Use property promotion

jackbayliss:
Noticed property promotion could be used in quite a few places.

Felt like a nice PR to keep things modern and align with the way framework is going imo.

No breaking changes..

Been a thing since PHP 8.0


jackbayliss:
Style is failing, but should be good to review - presume it'll fix itself if merged.

Issue #588, opened 2026-03-20
'Illuminate\Foundation\Http\Middleware\ValidateCsrfToken' is deprecated


SantosVilanculos:
### Sanctum Version

4.3.1

### Laravel Version

13.1.1

### PHP Version

8.4.18

### Description

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/30841a4d-b72d-406b-9d85-9d5038badc4b" />

### Steps To Reproduce

```sh
php artisan install:api
```

PR #589, opened 2026-03-20
[4.x] Rename CSRF middleware to PreventRequestForgery

Shuudy:
## Summary

This PR updates the CSRF middleware usage to align with `Laravel 13`.

The middleware has been renamed from `VerifyCsrfToken` and `ValidateCsrfToken` to `PreventRequestForgery`, and the previous versions are now deprecated. 
The current configuration was still using the old middleware, which caused deprecation warnings in the IDE.

This change was made following issue #588, which highlighted this deprecation.

## Breaking Changes

The configuration key has been renamed from `validate_csrf_token` to `prevent_request_forgery` to stay consistent with the new middleware naming.

### DIFF
--- a/config/sanctum.php
+++ b/config/sanctum.php
@@ -78,7 +78,7 @@ return [
     'middleware' => [
         'authenticate_session' => Laravel\Sanctum\Http\Middleware\AuthenticateSession::class,
         'encrypt_cookies' => Illuminate\Cookie\Middleware\EncryptCookies::class,
-        'validate_csrf_token' => Illuminate\Foundation\Http\Middleware\ValidateCsrfToken::class,
+        'prevent_request_forgery' => Illuminate\Foundation\Http\Middleware\PreventRequestForgery::class,
     ],
 
 ];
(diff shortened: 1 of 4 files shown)


config/sanctum.php
```diff
@@ -78,7 +78,7 @@
     'middleware' => [
         'authenticate_session' => Laravel\Sanctum\Http\Middleware\AuthenticateSession::class,
         'encrypt_cookies' => Illuminate\Cookie\Middleware\EncryptCookies::class,
-        'validate_csrf_token' => Illuminate\Foundation\Http\Middleware\ValidateCsrfToken::class,
+        'prevent_request_forgery' => Illuminate\Foundation\Http\Middleware\PreventRequestForgery::class,
```
crynobone, line 81: This will failed when used in Laravel 12 and below.

crynobone (changes requested):

config/sanctum.php
Shuudy: Thanks @crynobone, you’re right, this solution only works in Laravel 13, since this class exists only in that version.

To make it compatible across all versions, we would need to use something like `class_exists()` to dynamically select the correct class.

Then we could use the correct class, though I’m not entirely sure how to go about it.

I think something like this should do the trick: 

```php
class_exists(
    \Illuminate\Foundation\Http\Middleware\PreventRequestForgery::class
)
    ? \Illuminate\Foundation\Http\Middleware\PreventRequestForgery::class
    : \Illuminate\Foundation\Http\Middleware\ValidateCsrfToken::class;
```

However, in this specific case, this approach does not solve the fact that the old classes appear as deprecated.

SkandarS0:
I think a new major version would be a solution no ?

alexkart:
BTW: It looks like the Sanctum documentation (https://laravel.com/docs/13.x/sanctum#csrf-protection) needs to be updated, as it still states that sending the X-XSRF-TOKEN header is required and does not mention the new Sec-Fetch-Site origin shortcut.

PR #604, opened 2026-05-26
Change separator from `|` to `.` for tokens to be more compliant with advanced standards like OAuth

maximal:
I understand that Sanctum is not OAuth provider itself, but changing separators from pipes to dots would make tokens consumable by some other libraries in different languages (Kotlin, etc.), which now treat Sanctum’s tokens as invalid by default because of `|` separators.

See RFC 6750: https://datatracker.ietf.org/doc/html/rfc6750#section-2.1

### DIFF
--- a/src/Guard.php
+++ b/src/Guard.php
@@ -99,11 +99,11 @@ class Guard
      */
     protected function isValidBearerToken(?string $token = null)
     {
-        if (! is_null($token) && str_contains($token, '|')) {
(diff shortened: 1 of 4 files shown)


taylorotwell:

maximal:
@taylorotwell, I guess the amount of code stays the same, but in addition, we become more compliant with existing standards and RFCs.

PR #617, opened 2026-09-08
Update EnsureFrontendRequestsAreStateful.php

neomasterr:
VerifyCsrfToken is deprecated, use PreventRequestForgery instead


neomasterr:
didn't see 11 version support

